PR #13, opened 2022-03-09
Add note on extend data matrix.

aterentic-ethernal:


sh3ll3x3c (approved):
Just as a note, extension factor other than 2 has been tried, but reconstruction failed...

prabal-banerjee:
@sh3ll3x3c Is there a test which fails with extension factor above 2? I want to understand why it fails.

aterentic-ethernal:
@prabal-banerjee , lets add that test 👍

PR #16, opened 2022-03-16
Ava 90/bulk tx

kroos47:
- added bulk-tx script 

- run using `ts-node src/bulk.ts `or `npm run bulk`

### DIFF
--- a/tests/bulk-tx/src/bulk_tx.ts
+++ b/tests/bulk-tx/src/bulk_tx.ts
@@ -1,0 +1,141 @@
+import { ApiPromise, WsProvider, Keyring } from '@polkadot/api';
+import { SignerResult } from '@polkadot/api/types';
+import type { EventRecord, ExtrinsicStatus, H256, SignedBlock } from '@polkadot/types/interfaces';
+import type { ISubmittableResult ,SignatureOptions } from '@polkadot/types/types';
+
+const ALICE = '5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY';
+const HERMEZ = '5CAezZ6kHGHY2ZUVvpqpJ9v7xSPm5Xhdj2F7H7k2iKyRzRDH';
+const herMnemonic = 'very noise grant jealous elegant update famous own pulp tongue soccer true';
+
+const keyring = new Keyring({ type: 'sr25519' });
+
+const batch = 3;
+const size = 100;
+
+async function createApi():Promise<ApiPromise> {
+  // Initialise the provider to connect to the local node
+  // const provider = new WsProvider('wss://polygon-da-explorer.matic.today/ws');
+  const provider = new WsProvider('ws://127.0.0.1:9944');
+
+  // Create the API and wait until ready
+  return ApiPromise.create({ 
+    provider,
+    types: {
+        DataLookup: {
+          size: 'u32',
+          index: 'Vec<(u32,u32)>'
+        },
+        KateExtrinsicRoot: {
+          hash: 'Hash',
+          commitment: 'Vec<u8>',
+          rows: 'u16',
+          cols: 'u16'
+        },
+        KateHeader: {
+          parentHash: 'Hash',
+          number: 'Compact<BlockNumber>',
+          stateRoot: 'Hash',
+          extrinsicsRoot: 'KateExtrinsicRoot',
+          digest: 'Digest',
+          appDataLookup: 'DataLookup'
+        },
+        Header: 'KateHeader',
+        AppId: 'u32',
+        // CheckAppId: {
+        //     extra: {
+        //         appId: 'u32', 
+        //     },
+        //     types: {}
+        // }
+    },
+    signedExtensions: {
+      CheckAppId: {
+        extrinsic: {
+          appId: 'u32'
+        },
+        payload: {}
+      },
+    },
+  });
+}
+
+function generateData(size:number):string {
+    let randomChars = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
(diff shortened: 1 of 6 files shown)


tests/bulk-tx/src/bulk_tx.ts
```diff
@@ -0,0 +1,145 @@
...
+        },
+        payload: {}
+      },
+    },
+  });
+}
+
+//generating random data as per payload specified
+function generateData(size: number): string {
+  let randomChars = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
```
prabal-banerjee, line 62: Any reason to use alphanumeric? If not, then let's use random binary data.

tests/bulk-tx/src/bulk_tx.ts
```diff
@@ -0,0 +1,145 @@
...
+}
+
+interface SignatureOptionsNew extends SignatureOptions {
+  app_id: number
+}
+
+async function sendTx(api: ApiPromise, sender: any, nonce: any): Promise<any> {
+  try {
+    let data = generateData(size);
+    const nonc = await api.rpc.system.accountNextIndex(sender.address);
```
prabal-banerjee, line 82: To safely send multiple concurrent tx, we should probably manage nonce on our own.

tests/bulk-tx/src/bulk_tx.ts
```diff
@@ -0,0 +1,145 @@
...
+    let submit = await api.tx.dataAvailability.submitData(data);
+    const options: Partial<any> = { app_id: 1, nonce: -1 }
```
prabal-banerjee, line 84: We have `nonce` as an arg, `nonc` is the fetched nonce, and then send a constant nonce here. Confused about the exact approach.

tests/bulk-tx/src/bulk_tx.ts
```diff
@@ -0,0 +1,145 @@
...
+import { SignerResult } from '@polkadot/api/types';
+import type { EventRecord, ExtrinsicStatus, H256, SignedBlock } from '@polkadot/types/interfaces';
+import type { ISubmittableResult, SignatureOptions } from '@polkadot/types/types';
+
+const ALICE = '5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY';
+const keyring = new Keyring({ type: 'sr25519' });
+
+//batch size and payload size are set here
+const batch = 3;
+const size = 100;
```
prabal-banerjee, line 11: unit?

tests/bulk-tx/src/bulk_tx.ts
kroos47: Done

tests/bulk-tx/src/bulk_tx.ts
kroos47: Fixed.

tests/bulk-tx/src/bulk_tx.ts
```diff
@@ -0,0 +1,144 @@
...
+        },
+        payload: {}
+      },
+    },
+  });
+}
+
+//generating random data as per payload specified
+function randomDigit() {
+  return Math.floor(Math.random() * Math.floor(2));
```
prabal-banerjee, line 63: Why are we using `Math.floor(2)`?

tests/bulk-tx/src/bulk_tx.ts
```diff
@@ -0,0 +1,144 @@
...
+import { KeyringPair } from '@polkadot/keyring/types';
+import type { EventRecord, ExtrinsicStatus, H256, SignedBlock } from '@polkadot/types/interfaces';
+import type { ISubmittableResult, SignatureOptions } from '@polkadot/types/types';
+
+const ALICE = '5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY';
+const keyring = new Keyring({ type: 'sr25519' });
+
+//batch size and payload size are set here
+const batch:number = 3;
+const size:number = 100;
```
prabal-banerjee, line 12: Not sure in what unit the size is defined. bits, bytes, KB, MB, .. ?

prabal-banerjee (approved):

tests/bulk-tx/src/bulk_tx.ts
kroos47: To get the numbers 0 & 1. So effectively its like %2 to every number we multiplied with `math.random()`

tests/bulk-tx/src/bulk_tx.ts
kroos47: bytes. So if its 100, then its 100bytes

MiguelDD1 (approved):
It would be nice if `size` of the data and `batch` would be parameters from CLI.
We are duplicating some code between `tests`, I think we could unify these tests into one single tool and remove that duplication. Something like:
```
avail-client  <command>

COMMANDS:
   bulk-tx    Submit data in batch mode.
   get-tx-info   Get TX info for a given block.
   ...
   
```

tests/bulk-tx/src/bulk_tx.ts
prabal-banerjee: I understand that. But `Math.floor(2)` equals `2` always! Is there a special reason to use the floor function?

kroos47: So actually the result is different when using simple 2 and math.floor(2) here. Thats what I found

kroos47: oh nvm mind! Its the same! My mistake on it 😅

kroos47:
Reopened a simlar PR for merging the script

PR #18, opened 2022-03-17
add:scripts merged to a single one

kroos47:
Script which contains the merged script to send,retrieve or bulk-tx

- Run the script using `ts-node/tests/test-scripts/src/full.ts` or using `npm run start --` 
with optional arguments: 
  -n "bulk_tx"/"app"/"get_data" where `bulk_tx` is used to send bulk tx, `app` is used to send a single extrinsic data and `get_data` is to send a single extrinsic data and get that extrinsic
  -b is the batch which is the number of tx to be sent
  -s is the payload of the tx in bytes
  -id is the app_id value
  -e is the endpoint taken as a string


tests/test-scripts/src/full.ts
```diff
@@ -0,0 +1,225 @@
...
+            default: 'app'
+        }
+
+    }).argv;
+}
+
+async function createApi(): Promise<ApiPromise> {
+    // Initialise the provider to connect to the local node
+    // const provider = new WsProvider('wss://polygon-da-explorer.matic.today/ws');
+    const provider = new WsProvider('ws://127.0.0.1:9944');
```
prabal-banerjee, line 46: Shouldn't this be fetched from `argv`?

tests/test-scripts/src/full.ts
```diff
@@ -0,0 +1,225 @@
...
+                parentHash: 'Hash',
+                number: 'Compact<BlockNumber>',
+                stateRoot: 'Hash',
+                extrinsicsRoot: 'KateExtrinsicRoot',
+                digest: 'Digest',
+                appDataLookup: 'DataLookup'
+            },
+            Header: 'KateHeader',
+            AppId: 'u32',
+            // CheckAppId: {
```
prabal-banerjee, line 72: Let us remove it if it is no longer relevant

tests/test-scripts/src/full.ts
```diff
@@ -0,0 +1,225 @@
...
+    return yargs(process.argv.slice(2)).options({
+        e: {
+            description: 'WSS endpoint',
+            alias: 'endpoint',
+            type: 'string',
+            default: 'wss://polygon-da-explorer.matic.today/ws'
+        },
+
+        s: {
+            description: 'payload to be given in bytes',
```
prabal-banerjee, line 20: Are we sure this is in bytes?

tests/test-scripts/src/full.ts
```diff
@@ -0,0 +1,225 @@
...
+
+async function sendTx(api: ApiPromise, sender: KeyringPair, nonce: number, argv: any): Promise<any> {
+    try {
+
+        let payload = argv.s;
+        let data = generateRandomBinary(payload);
+        let submit = await api.tx.dataAvailability.submitData(data);
+        /* @note here app_id is 1,
+        but if you want to have one your own then create one first before initialising here */
+        const options: Partial<any> = { app_id: 1, nonce: nonce }
```
prabal-banerjee, line 120: Can we take `appID` also as cmdline arg?

prabal-banerjee (approved):

tests/test-scripts/src/full.ts
kroos47: yes that's what I got when I checked

PR #20, opened 2022-03-21
Miguel/ava 33 create monk template for da testnet

MiguelDD1:
- [x] Add DevNet Chain-spec
- [x] Some corrections on Testnet:
  - Security: Controller accounts are segregated from Stash accounts
  - Session keys are fixed.


aterentic-ethernal:
Should we move devnet and testnet chainspec to avail-deployment or that makes no sense?

kroos47:
> Should we move devnet and testnet chainspec to avail-deployment or that makes no sense?

I think It was in that manner before but changed because of the need to copy the file time to time from local to docker. @MiguelDD1

MiguelDD1:
>    >   Should we move devnet and testnet chainspec to avail-deployment or that makes no sense?

> I think It was in that manner before but changed because of the need to copy the file time to time from local to docker. @MiguelDD1

Yes that was the problem. In this repo we will get at least 3 chain specs: DevNet, TestNet, and MainNet.
In any case, the `--chain` parameter allows any kind of customization on the deployment.

kroos47 (approved):

PR #30, opened 2022-04-05
Add proof testing to reconstruction test.

aterentic-ethernal:
This PR adds proof generation and verification to test_build_and_reconstruct. Proof is generated and verified for 1% cells of 20 randomly generated inputs as temporary performance constraint. This should fine starting point to cover most cases in complete encode/reconstruct/generate and verify proof flow. Some of the cases are not covered, like when proof is incorrect and that should be added as separate test later.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2929,6 +2929,7 @@ dependencies = [
  "frame-support",
  "getrandom 0.2.4",
  "hex",
+ "kate-proof",
  "kate-recovery",
  "log",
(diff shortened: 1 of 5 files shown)


sh3ll3x3c:
NOTE: Proptest(s) will generate testcase files when failing. As a good practice, those files should be commited (not ignored) as the future proptests will use them to make sure the bug is indeed fixed...

prabal-banerjee (approved):

PR #33, opened 2022-04-07
Extract kate-rpc to a separate crate

sh3ll3x3c:
- kate-rpc logic has been extracted from data-avail (node) crate to a separate crate inside. 

The reasoning behind this move, other than pure code organization, is to enable a faster code instrumentation for the `query_proof`  function, as instrumenting the entire node is currently impossible due to WASM issues with AFL. 

TODO: remove unused deps from the data-avail crate


sh3ll3x3c:
NOTE: the `runtime` crate was creating compile issues when moving `kate-rpc-runtime-api` and `kate-rpc` into a single crate.

aterentic-ethernal (approved):

prabal-banerjee (approved):
lgtm. but would like @MiguelDD1 to approve it before merge.

MiguelDD1 (approved):

PR #38, opened 2022-04-20
Ava 120

aterentic-ethernal:


kate/recovery/Cargo.toml
```diff
@@ -5,6 +5,7 @@ authors = ["Denis Ermolin <denis.ermolin@matic.network>"]
 edition = "2018"
 
 [dependencies]
+anyhow = "1.0.41"
```
MiguelDD1, line 8: Can we replace `anyhow` by `thiserror`?
`thiserror` is better for libraries, and allows you to make a proper error handling when you use the library.

kate/recovery/src/com.rs
```diff
@@ -36,46 +37,40 @@ pub fn unflatten_padded_data(
 	layout: Vec<(u32, u32)>,
 	data: Vec<u8>,
 	chunk_size: usize,
-) -> Vec<(u32, Vec<u8>)> {
-	assert!(data.len() % chunk_size == 0);
-	let data = data
+) -> Result<Vec<(u32, Vec<u8>)>> {
```
MiguelDD1, line 40: That could be an issue because `unfatten_padded_data` is used in `System::finalize` which should be **infallible**.

kate/recovery/Cargo.toml
aterentic-ethernal: Done.

sh3ll3x3c (approved):

PR #41, opened 2022-04-21
Miguel/ava 152 BlockTooBig bug

MiguelDD1:
# Bug Description
During the finalization phase of the block creation, we build the kate commitment. This process MUST be infallible, so the preconditions from kate commitment MUST be double-checked before add any TX into the execution queue.
If the kate commitment fails, and the block creation panics and the TX will be included in the next try of block baking. As a result, the entire network will be stuck indefinitely.

# How to reproduce
Schedule a TX using the schedule pallet to execute a Code Runtime Upgrade and a payload greater than 2.6MB.

# Changes
- [x] The _signed extension_ `check_weight` now double-check the **Padded Block Len Limit** too. It means, TX will NOT be added into a block if 
- [x] Add **static** and **debug assertions** on `chunk_size`. It must be greater that `DATA_CHUNK_SIZE`.
- [x] New functions to pre-calculate the padded block len, and UTs to verify they are aligned with actual functions.
- [x] `System::DynamicBlockLength` replaces the raw access to storage of `System::set_block_length`.
- [x] `System::AllExtrinsicsLen` also tracks the padded len.
- [x] New migrations for `System`. Pallet version = 1

### DIFF
--- a/pallets/system/src/lib.rs
+++ b/pallets/system/src/lib.rs
@@ -108,6 +108,7 @@ pub mod offchain;
 
 mod extensions;
 pub mod header_builder;
+pub mod migrations;
 use header_builder::HeaderBuilder;
 
 #[cfg(feature = "std")]
@@ -330,14 +331,7 @@ pub mod pallet {
 
 	#[pallet::hooks]
 	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
-		fn on_runtime_upgrade() -> frame_support::weights::Weight {
-			if !UpgradedToTripleRefCount::<T>::get() {
-				UpgradedToTripleRefCount::<T>::put(true);
-				migrations::migrate_to_triple_ref_count::<T>()
-			} else {
-				0
-			}
-		}
+		fn on_runtime_upgrade() -> frame_support::weights::Weight { migrations::migrate::<T>() }
 
 		fn integrity_test() {
 			T::BlockWeights::get()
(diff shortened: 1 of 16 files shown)


pallets/system/src/lib.rs
```diff
@@ -1428,7 +1456,12 @@ impl<T: Config> Pallet<T> {
 		BlockWeight::<T>::mutate(|current_weight| {
 			current_weight.set(weight, DispatchClass::Normal)
 		});
-		AllExtrinsicsLen::<T>::put(len as u32);
+		let block_length = Self::block_length();
+		let padded = kate::padded_len(len as u32, block_length.chunk_size());
```
sh3ll3x3c, line 1460: @MiguelDD1 might be a naive question, but why do we need the padded length here, if the padded block size is only used for calculating commitments?

MiguelDD1: This is due to a security issue. 
We update the value here and **specially** at `check_weight.rs` (fn `do_pre_dispatch`). This latest one is called before every TX is executed, so It allows us to filter TXs just before add it to the block, and it means we are sure that `fn finalize()` is infallible.
That was the root-cause of  `BlockTooBig` issue: `fn finalize` MUST be infallible but the padding len limit was checked inside, having no way to recovery neither to pre-filter TX.

prabal-banerjee (approved):

kroos47 (approved):
so we check for weight and block padded size if the block size is bigger right?Just to get an understaning?

PR #44, opened 2022-05-11
Miguel/inf2 298 explorer cannot decode some extrinsics

MiguelDD1:
# Description
It adds a new UT to double-check that extrinsics decodification is working on the node.

# Extra
- [x] Replaces `unsafe` code by equivalent `safe` code.
- [x] Adds static assertion to ensure the linking between `PROVER_KEY_SIZE` and `G1Affine::SIZE` which is the output of `Commitment::to_bytes()`

### DIFF
--- a/kate/src/com.rs
+++ b/kate/src/com.rs
@@ -7,6 +7,7 @@ use std::{
 use da_primitives::asdr::AppExtrinsic;
 use dusk_bytes::Serializable;
 use dusk_plonk::{
+	bls12_381::G1Affine,
 	commitment_scheme::kzg10,
 	error::Error as PlonkError,
 	fft::{EvaluationDomain, Evaluations},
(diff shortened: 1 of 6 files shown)


kroos47 (approved):

kate/src/com.rs
```diff
@@ -383,11 +384,9 @@ pub fn build_commitments(
 	let (prover_key, _) = public_params.trim(block_dims.cols).map_err(Error::from)?;
 	let row_eval_domain = EvaluationDomain::new(block_dims.cols).map_err(Error::from)?;
 
-	let mut result_bytes: Vec<u8> = Vec::new();
-	result_bytes.reserve_exact(PROVER_KEY_SIZE * extended_rows_num);
-	unsafe {
-		result_bytes.set_len(PROVER_KEY_SIZE * extended_rows_num);
-	}
+	// @internal: `PROVER_KEY_SIZE` must match with the output size of `Commitment::to_bytes`.
+	const_assert_eq!(G1Affine::SIZE, PROVER_KEY_SIZE);
```
prabal-banerjee, line 388: In which circumstances can this assertion fail? If the config is wrong, the assertion will fail but should we handle that case more gracefully instead of shutting down the node?

prabal-banerjee (approved):

kate/src/com.rs
MiguelDD1: This assertion is verified at **compilation time**, so if both constants are different the _node will not compile_. See [static_assertions crate](https://docs.rs/static_assertions/1.1.0/static_assertions/index.html)
It has 0 overhead  at runtime.

prabal-banerjee: Thanks! Didn't know that.

PR #51, opened 2022-05-31
Adding the Testnet Chainspec

praetoriansentry:
Adding the testnet-v2 chainspec which is being used for testnet.polygonavail.net


vcastellm (commented):
Late LGTM

PR #54, opened 2022-06-08
Luka/ava 208 testnet stuck

luka-ethernal:


pallets/system/src/header_builder.rs
```diff
@@ -106,13 +106,15 @@ pub trait HostedHeaderBuilder {
 
 		use crate::generic::DigestItem;
 
+		let mut sorted_app_extrinsics = app_extrinsics.clone();
+		sorted_app_extrinsics.sort_by(|a, b| a.app_id.cmp(&b.app_id));
```
aterentic-ethernal, line 110: Just random thought, maybe we could move sorting into build commitments function to avoid depending on sorted extrinsics, and also maybe reuse it in rpc 🤷‍♂️

luka-ethernal: That's a great idea! Good thinking, I would've never thought of it.

luka-ethernal: On second thought, although it could reduce repetition, perhaps semantically, it is better this way. Let's discuss tomorrow.

aterentic-ethernal (approved):

prabal-banerjee (approved):

PR #59, opened 2022-06-29
Remove sp-core dependency from kate/proof.

aterentic-ethernal:
This PR removes sp-core dependency from kate/proof to decouple avail-light from sp-core version defined in avail project.

### DIFF
--- a/kate/proof/src/lib.rs
+++ b/kate/proof/src/lib.rs
@@ -22,8 +22,7 @@ mod testnet {
 
 pub struct ProofVerification {
 	pub status: Result<(), dusk_plonk::error::Error>,
-	pub public_params_hash: String,
-	pub public_params_len: usize,
+	pub public_params: Vec<u8>,
 }
 
 // code for light client to verify incoming kate proofs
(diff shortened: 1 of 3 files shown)


luka-ethernal (approved):

luka-ethernal (approved):

kroos47 (approved):

kate/proof/src/lib.rs
```diff
@@ -83,7 +82,6 @@ pub fn kc_verify_proof(
 
 	Ok(ProofVerification {
 		status,
-		public_params_hash: hex::encode(sp_core::blake2_128(&raw_pp)),
-		public_params_len: hex::encode(raw_pp).len(),
+		public_params: raw_pp,
```
prabal-banerjee, line 85: Should we keep a log of hash+len of public params in debug log level? Might be useful. We need not use `blake2`. Any standard hash function like SHA256 or even MD5 might suffice.

prabal-banerjee (approved):

sh3ll3x3c (approved):

PR #60, opened 2022-07-01
Describe runtime upgrade process in README.md.

aterentic-ethernal:


luka-ethernal (approved):

README.md
```diff
@@ -122,3 +122,33 @@ To clean up generate coverage information files, run
...
+If needed, use `wasm-opt` from [Binaryen](https://github.com/WebAssembly/binaryen) to optimize size of a WASM runtime:
+
+	$> wasm-opt -Oz -o ./da_runtime.compact.wasm \
+		./runtime/target/srtool/release/wbuild/da-runtime/da_runtime.compact.wasm
+
+### Upgrade process
+
+Since we have block size and transaction weight limits, upgrade is a three step process:
+
+1. Use `sudo/sudoCall` to invoke `dataAvailability/submit_block_length_proposal` with increased block limits
```
prabal-banerjee, line 145: In docs, we should not recommend `sudo` calls. IMO, we should have a way to run upgrade using democracy pallet, which internally calls the above `sudo` calls. This will ensure that future upgrades can be done in a decentralized manner and can be done on canary/mainnet. 

We can keep these as part of devnet upgrades.

kroos47 (approved):

PR #61, opened 2022-07-06
Proof verification change(including anjan's benchmarking) 

kroos47:


aterentic-ethernal:
@kroos47 , why do we want to merge both features in one PR? Also, it would be great if we could abstract differences of build_commitments, par_build_commitments and opt_par_build_commitments since there are a lot of repeated code, or remove versions which are not optimal, if possible. Not sure about unsafe code also, is there some alternative to explicitly fill chunks of a vector in parallel, like collecting results and concatenating them?

aterentic-ethernal:
Another thing, do we want to remove unwraps or its ok to leave as it is? This would require little more effort but it will make node more resilient with clear error messages. At least we shouldn't introduce new unwraps.

kroos47:
> @kroos47 , why do we want to merge both features in one PR? Also, it would be great if we could abstract differences of build_commitments, par_build_commitments and opt_par_build_commitments since there are a lot of repeated code, or remove versions which are not optimal, if possible. Not sure about unsafe code also, is there some alternative to explicitly fill chunks of a vector in parallel, like collecting results and concatenating them?

So we can do in two ways as you said one as one PR with this only change and one PR for benchmarking. But I thought we can check one thing within the same thing. 

Also I will update the recurssive codes from the functions.

kroos47:
Closing PR to create a new one with different PRs one with anjan's change(bench and kate) and one with proof function change only

PR #62, opened 2022-07-11
Update runtime upgrade description to use democracy instead of sudo.

aterentic-ethernal:


prabal-banerjee (commented):
Can we keep both the ways for future reference? As mentioned, one will be preferred way and another will be the `sudo` approach.

prabal-banerjee (approved):

PR #74, opened 2022-08-31
Add data root field inside header extrinsic root

prabal-banerjee:


pallets/system/src/header_builder.rs
```diff
@@ -158,3 +180,86 @@ pub trait HostedHeaderBuilder {
 		)
 	}
 }
+
+#[derive(Debug, Clone, PartialEq, Eq, Default)]
+pub struct AvailExtrinsic {
```
aterentic-ethernal, line 185: Maybe it makes sense to move this struct to da-primitives and reuse it in both header builder and rpc call?

aterentic-ethernal (approved):
Looks good to me, just formatting and if its possible, we should reuse structs.

aterentic-ethernal (changes requested):
There is also a test in runtime/src/lib.rs which fails to compile, so changes has to be made before merge

kroos47:
> There is also a test in runtime/src/lib.rs which fails to compile, so changes has to be made before merge

yeh actually multiple other tests needs to be fixed, so working on it. 😄

kroos47 (approved):

PR #80, opened 2022-09-13
Exposes `Header::Number` as compact type

MiguelDD1:
It solves the decoding issue of the header between `subxt` & the node.

**NOTE**:
It requires a complete clean (`cargo clean`) on `avail-subxt` due to changes on `avail.dev.metadata.scale`.

### DIFF
--- a/primitives/src/header.rs
+++ b/primitives/src/header.rs
@@ -76,6 +76,7 @@ pub struct Header<Number: HeaderNumberTrait, Hash: KateHashTrait> {
 	pub parent_hash: Hash::Output,
 	/// The block number.
 	#[cfg_attr(feature = "std", serde(with = "number_serde"))]
+	#[codec(compact)]
 	pub number: Number,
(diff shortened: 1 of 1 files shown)


kroos47 (approved):

prabal-banerjee (approved):

primitives/src/header.rs
```diff
@@ -145,13 +146,20 @@ where
 	Hash::Output: Decode,
 {
 	fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
+		let parent_hash = Decode::decode(input)?;
```
aterentic-ethernal, line 149: Curios about extracting inlined code to vars. Is reasoning to ease debugging?

aterentic-ethernal (approved):

primitives/src/header.rs
MiguelDD1: Yes, just easy debugging.

PR #82, opened 2022-09-19
add tests

kroos47:


pallets/system/src/header_builder.rs
```diff
@@ -263,3 +264,207 @@ impl Decode for AvailExtrinsic {
...
+	};
+	data_root
+}
+
+mod tests {
+	use super::*;
+
+	/// tests for data root is created correctly
+	#[test]
+	fn data_root_test() {
```
aterentic-ethernal, line 289: Probably we can use hex and test_case macros to enumerate cases and make them clearer and shorter.

pallets/system/src/header_builder.rs
```diff
@@ -125,24 +125,25 @@ pub trait HostedHeaderBuilder {
...
-				.map(|x| Sha256::hash(&x.data))
-				.collect();
-
-			let data_tree = MerkleTree::<Sha256>::from_leaves(&leaves);
-			data_tree.root().expect("Data Root computation failed")
-		} else {
-			Default::default()
-		};
+		let extrinsics: Vec<Vec<u8>> = app_extrinsics.clone().into_iter().map(|e| e.data).collect();
+		// let avail_extrinsics = extrinsics
```
prabal-banerjee, line 129: remove comments?

pallets/system/src/header_builder.rs
```diff
@@ -125,24 +125,25 @@ pub trait HostedHeaderBuilder {
...
+		// 		.iter()
+		// 		.map(|x| Sha256::hash(&x.data))
+		// 		.collect();
+
+		// 	let data_tree = MerkleTree::<Sha256>::from_leaves(&leaves);
+		// 	data_tree.root().expect("Data Root computation failed")
+		// } else {
+		// 	Default::default()
+		// };
+		let data_root = data_root_avail(app_extrinsics);
```
prabal-banerjee, line 146: nit: can we name the func better? like `build_data_root` or something? `Avail` is implicit so we can drop it from most references.

pallets/system/src/header_builder.rs
```diff
@@ -263,3 +264,207 @@ impl Decode for AvailExtrinsic {
...
+		let tree = MerkleTree::<Sha256>::from_leaves(&leaves);
+		tree.root().expect("Data Root computation failed")
+	} else {
+		Default::default()
+	};
+	data_root
+}
+
+mod tests {
+	use super::*;
```
prabal-banerjee, line 285: GitHub action says this is unused import. Kindly recheck.

pallets/system/src/header_builder.rs
```diff
@@ -263,3 +264,207 @@ impl Decode for AvailExtrinsic {
...
+			.to_vec(),
+		];
+
+		let leaves = avail_data
+			.iter()
+			.map(|xt| Sha256::hash(&xt))
+			.collect::<Vec<[u8; 32]>>();
+
+		let merkle_tree = MerkleTree::<Sha256>::from_leaves(&leaves);
+		let indices_to_prove = vec![0, 1];
```
prabal-banerjee, line 459: this probably checks a multi-leaf proof. do we expose that functionality through RPC?

pallets/system/src/header_builder.rs
```diff
@@ -263,3 +264,207 @@ impl Decode for AvailExtrinsic {
...
+					248, 62, 54, 10, 93, 148, 47, 231, 94, 157, 88, 228, 145, 6, 168, 232, 178, 54,
+					1, 203, 198, 166, 51, 216, 14, 93, 8, 157, 131, 164, 0, 4, 0, 3, 0, 0, 0, 29,
+					1, 164, 104, 104, 97, 106, 107, 97, 110, 99, 107, 97, 32, 108, 97, 104, 32,
+					105, 97, 107, 106, 32, 99, 97, 105, 117, 104, 32, 97, 98, 32, 97, 105, 97, 106,
+					104, 32, 97, 32, 103, 97, 104, 97,
+				],
+			},
+		];
+		let empty_data: Vec<AppExtrinsic> = vec![];
+		let single_data: Vec<AppExtrinsic> = vec![AppExtrinsic {
```
prabal-banerjee, line 311: so this extrinsic is not a `submitData()` call?

pallets/system/src/header_builder.rs
```diff
@@ -263,3 +264,207 @@ impl Decode for AvailExtrinsic {
...
+			[
+				50, 98, 49, 49, 102, 49, 100, 100, 51, 57, 53, 53, 54, 102, 98, 50, 97, 98, 52, 50,
+			]
+			.to_vec(),
+		];
+
+		for (pos, xt) in avail_data.iter().enumerate() {
+			leaves.push(Sha256::hash(&xt));
+
+			if pos == 1usize {
```
prabal-banerjee, line 354: We already know the position. Do we need to put this inside the loop?

pallets/system/src/header_builder.rs
```diff
@@ -263,3 +264,207 @@ impl Decode for AvailExtrinsic {
...
+				interested_leaf_position = Some(leaves.len() - 1);
+			}
+		}
+
+		if leaves.len() > 0 {
```
prabal-banerjee, line 359: This is a static proof value assertion. Why do we need to mock the construction logic here?

pallets/system/src/header_builder.rs
```diff
@@ -263,3 +264,207 @@ impl Decode for AvailExtrinsic {
...
+
+		let proof = merkle_tree.proof(&indices_to_prove);
+		let root = merkle_tree
+			.root()
+			.ok_or("couldn't get the merkle root")
+			.unwrap();
+
+		assert!(proof.verify(root, &indices_to_prove, leaves_to_prove, leaves.len()));
+	}
+}
```
prabal-banerjee, line 470: Can we add cases where we ask for proof when: 
- tree has no leaf
- tree has single leaf
- tree has power of 2 leaves

pallets/system/src/header_builder.rs
```diff
@@ -263,3 +264,207 @@ impl Decode for AvailExtrinsic {
...
+				let proof = data_tree.proof(&[position]);
+				let root_proof = proof.proof_hashes().to_vec();
+				// here the proof is shown empty because the root itself is the proof as there is only one appdata extrinsic
+				assert_eq!(root_proof, empty_vec);
+			}
+		}
+	}
+
+	#[test]
+	fn test_no_data_merkle_proof() {
```
aterentic-ethernal, line 408: I'm not sure this test depends on any external code, it looks like it could be removed. Maybe we are missing some higher level abstraction, since part of a code is repeated in tests (`if leaves.len() > 0...`)

pallets/system/src/header_builder.rs
kroos47: yes it is a submitdata() but here only the decoded data is directly given into the function

prabal-banerjee: Then how is the root empty in this case?

pallets/system/src/header_builder.rs
kroos47: no but I tried it actually when I saw this in the crate! can remove it if not needed

pallets/system/src/header_builder.rs
kroos47: yes

pallets/system/src/header_builder.rs
kroos47: yeh but it actually is needed

pallets/system/src/header_builder.rs
```diff
@@ -263,3 +247,178 @@ impl Decode for AvailExtrinsic {
...
+		.collect::<Vec<_>>();
+	let data_root: [u8; 32] = if avail_ext.len() > 0 {
+		let leaves: Vec<[u8; 32]> = avail_ext.iter().map(|e| Sha256::hash(&e.data)).collect();
+		let tree = MerkleTree::<Sha256>::from_leaves(&leaves);
+		tree.root().expect("Data Root computation failed")
+	} else {
+		Default::default()
+	};
+	data_root
+}
```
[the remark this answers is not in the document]
prabal-banerjee: @MiguelDD1 We want the data root to be the Merkle root of only `SubmitData()` calls, i.e., the leaves of the tree should contain only the data posted by users and *not* 
- all the extrinsics in the block
- encoded extrinsics

This is because, on Ethereum (or destination chain), we want to give proofs of data inclusion. If we push `extrinsic_root()`, then
- Verifying proof of inclusion is costly (Blake2 is gas expensive)
- Even if we create SHA/keccak based merkle tree, still to show inclusion we need to
  - give the entire encoded extrinsic -> this defeats purpose of Validiums
  - verify proof of inclusion
  - decode the extrinsic -> gas expensive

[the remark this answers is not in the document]
prabal-banerjee: @kroos47 Let's change the root field to `H256` and update/test the LC, explorer and test scripts. Should work without much trouble IMO.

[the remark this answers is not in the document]
kroos47: done!

MiguelDD1 (approved):

PR #83, opened 2022-09-19
Post nomad primitives PR

MiguelDD1:
This PR contains minor improvements after Nomad primitives inclusion:

- Use `keccak256_concat` macro to **avoid extra allocations**.
- Replace `lazy_static` by a **pre-build value**. An unit test was added to ensure both are the same value.
- `NomadMessage::to_vec` uses less memory.
- New UT for `NON_BODY_LENGTH` to ensure it is synchronized with `NomadMessage`.

### DIFF
--- a/primitives/nomad/merkle/src/lib.rs
+++ b/primitives/nomad/merkle/src/lib.rs
@@ -31,20 +31,112 @@ pub type NomadLightMerkle = light::LightMerkle<TREE_DEPTH>;
 pub type NomadProof = proof::Proof<TREE_DEPTH>;
 
 pub use error::{TreeError, VerifyingError, VerifyingError::VerificationFailed};
+use hex_literal::hex;
 pub use light::*;
 pub use proof::*;
 pub use utils::*;
 
-lazy_static::lazy_static! {
-	/// A cache of the zero hashes for each layer of the tree.
-	pub static ref ZERO_HASHES: [H256; TREE_DEPTH + 1] = {
-		let mut hashes = [H256::zero(); TREE_DEPTH + 1];
-		for i in 0..TREE_DEPTH {
-			hashes[i + 1] = hash_concat(hashes[i], hashes[i]);
-		}
-		hashes
-	};
-}
+/// A cache of the zero hashes for each layer of the tree.
+/// See `zero_hashes_pre_build()` test to check how it is calculated.
+pub const ZERO_HASHES: [H256; TREE_DEPTH + 1] = [
+	H256::zero(),
+	H256(hex!(
+		"ad3228b676f7d3cd4284a5443f17f1962b36e491b30a40b2405849e597ba5fb5"
+	)),
+	H256(hex!(
+		"b4c11951957c6f8f642c4af61cd6b24640fec6dc7fc607ee8206a99e92410d30"
+	)),
+	H256(hex!(
+		"21ddb9a356815c3fac1026b6dec5df3124afbadb485c9ba5a3e3398a04b7ba85"
+	)),
+	H256(hex!(
+		"e58769b32a1beaf1ea27375a44095a0d1fb664ce2dd358e7fcbfb78c26a19344"
+	)),
+	H256(hex!(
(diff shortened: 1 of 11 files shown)


primitives/nomad/merkle/src/lib.rs
```diff
@@ -31,20 +31,112 @@ pub type NomadLightMerkle = light::LightMerkle<TREE_DEPTH>;
...
-		let mut hashes = [H256::zero(); TREE_DEPTH + 1];
-		for i in 0..TREE_DEPTH {
-			hashes[i + 1] = hash_concat(hashes[i], hashes[i]);
-		}
-		hashes
-	};
-}
+/// A cache of the zero hashes for each layer of the tree.
+/// See `zero_hashes_pre_build()` test to check how it is calculated.
+pub const ZERO_HASHES: [H256; TREE_DEPTH + 1] = [
```
aterentic-ethernal, line 41: Why we want to avoid lazy_static, is that something we should avoid? I'm not sure if it makes sense to change `TREE_DEPTH`, but in previous implementation you could change it without causing compilation to fail.

primitives/nomad/nomad-core/src/utils.rs
```diff
@@ -1,40 +1,15 @@
...
-pub fn keccak256<S>(bytes: S) -> [u8; 32]
-where
-	S: AsRef<[u8]>,
-{
-	let mut output = [0u8; 32];
-	let mut hasher = Keccak::v256();
-	hasher.update(bytes.as_ref());
-	hasher.finalize(&mut output);
-	output
+	keccak256_concat!(home_domain.to_be_bytes(), NOMAD_PREFIX)
```
aterentic-ethernal, line 8: Is nomad prefix or suffix? :)

primitives/nomad/merkle/src/lib.rs
MiguelDD1: The WASM interpreter could be re-loaded several times (i.e. after a `set_code` tx), so next call to `ZERO_HASHES` could spend more time than expected (a.k.a. calculated by benchmark) and the block execution could be out of time (the worst case), producing forks.

That situation will be really difficult to track/debug. So the idea is to replace any calculation at runtime by its output at compile time, specially if it is long time operations.

In order to warranty that pre-image contains the proper calculated values, we have the test case `zero_hashes_pre_build()`.

primitives/nomad/nomad-core/src/utils.rs
MiguelDD1: I guess It is suffix :rofl:

primitives/nomad/merkle/src/lib.rs
aterentic-ethernal: Oh, I didn't considered WASM runtime. Sounds like nice bug to have, though :D

primitives/nomad/nomad-core/src/lib.rs
```diff
@@ -1,5 +1,33 @@
...
+
+/// Calculates the Kecck 256 of arguments with NO extra allocations to join inputs.
+#[macro_export]
+macro_rules! keccak256_concat{
+	($($arg:tt)*) => {{
+		{
+			use tiny_keccak::Hasher as _;
+			let mut output = [0u8; 32];
+			let mut hasher = tiny_keccak::Keccak::v256();
+			$crate::keccak256_concat_update!(hasher, $($arg)*);
```
aterentic-ethernal, line 24: I have no experience with macros, but wouldn't `Vec` argument instead of vararg and for loop allow using this as fn instead of macro? If thats desired anyway :D

aterentic-ethernal (approved):

primitives/nomad/nomad-core/src/lib.rs
MiguelDD1: This macro has two main objectives:
- Reduce code duplication (which can also be achieve using `vec` as you suggested).
- Remove the extra memory allocation and/or extra copies. Using `vec` will require an extra heap allocation (required by `vec` itself), and deal with references and lifetimes to avoid copies between source objects into `vec`.

PR #84, opened 2022-09-19
Add nomad v1 pallets

luketchang:
Creates new `nomad` folder for all Nomad-related pallets:
  - `updater-manager`: handles the updating and slashing of the updater (largely unimplemented given that we haven't added updater slashing yet)
  - `home`: outbox for Avail which hosts a merkle tree of outbound messages + valid attestations from the updater entity
  - `da-bridge`: application responsible for validating block headers and enqueuing them as `DABridgeMessage`s to the home for dispatching to Ethereum


pallets/nomad/da-bridge/src/benchmarking.rs
```diff
@@ -0,0 +1,20 @@
+// //! Benchmarking setup for pallet-template
```
luketchang, line 1: Noting that all pallets are still missing benchmarks

node/src/chain_spec.rs
```diff
@@ -320,6 +320,21 @@ pub fn testnet_genesis(
...
+			_phantom: Default::default(),
+		},
+		home: HomeConfig {
+			local_domain: 2000,
+			committed_root: Default::default(),
+			updater: "0x1563915e194d8cfba1943570603f7606a3115508"
+				.parse()
+				.unwrap(),
+			_phantom: Default::default(),
+		},
```
luketchang, line 336: Addresses and domains for demo

Cargo.toml
```diff
@@ -12,6 +12,9 @@ members = [
 	"pallets/executive",
 	"pallets/system",
 	"pallets/dactr",
+	"pallets/nomad/updater-manager",
+	"pallets//nomad/home",
+	"pallets//nomad/da-bridge",
```
aterentic-ethernal, line 17: Are those `//` obsolete?

pallets/nomad/da-bridge/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+				body,
+			)?;
+
+			// Clear previous block_number to hash mappings, starting at the
+			// current submitted block_number going backwards. Because the
+			// runtime maps _every_ block number's hash, the sequence is always
+			// contiguous. If we are tracing backwards and find a block number
+			// that doesn't have a hash, we know everything before it has been
+			// cleared.
+			while FinalizedBlockNumberToBlockHash::<T>::contains_key(block_number) {
```
MiguelDD1, line 159: :bangbang: **Security** : Unbounded operation. It is impossible to determine the cost of `try_dispatch_data_root`.

pallets/nomad/da-bridge/src/message.rs
```diff
@@ -0,0 +1,109 @@
...
+	Invalid = 0,
+	DataRootMessage = 1,
+}
+
+impl From<u8> for DABridgeMessageTypes {
+	fn from(num: u8) -> Self {
+		match num {
+			0 => Self::Invalid,
+			1 => Self::DataRootMessage,
+			_ => panic!("Invalid u8 for DABridgeMessage enum!"),
```
MiguelDD1, line 19: :bangbang: **Security**: Remove `panic!` from production code. Replace by `TryFrom`

pallets/nomad/da-bridge/src/message.rs
```diff
@@ -0,0 +1,109 @@
...
+		}
+	}
+}
+
+/// Data root message variant. Contains a block number and its corresponding
+/// data root.
+#[derive(Clone, Debug, PartialEq)]
+pub struct DataRootMessage {
+	/// Block number
+	pub block_number: u32,
```
MiguelDD1, line 29: `block_number` should be a generic type, like `frame_system::Config::BlockNumber`.

pallets/nomad/home/src/lib.rs
```diff
@@ -0,0 +1,363 @@
...
+			message_body: BoundedVec<u8, T::MaxMessageBodyBytes>,
+		) -> DispatchResult {
+			Self::ensure_not_failed()?;
+
+			// Check message length against max
+			let message_length = message_body.len() as u32;
+			ensure!(
+				message_length < T::MaxMessageBodyBytes::get(),
+				Error::<T>::MessageTooLarge
+			);
```
MiguelDD1, line 209: Is not already checked by the type itself (as `BoundedVec<u8, T::MaxMessageBodyBytes>`)?
```suggestion
```

avail-subxt/examples/da_bridge_actor.rs
```diff
@@ -0,0 +1,57 @@
...
+
+use avail_subxt::*;
+use sp_keyring::AccountKeyring;
+use subxt::{ext::sp_core::H160, tx::PairSigner, OnlineClient};
+
+const DESTINATION_DOMAIN: u32 = 1000;
+const DA_BRIDGE_ROUTER_ADDRESS: &str = "0x77534486c6467fd24b1f7d60ca61d984d91f6a2a";
+
+#[async_std::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
```
aterentic-ethernal, line 11: Should we remove unwraps in while loop code to avoid confusion when someone blindly copy example? I've tried to improve it, not sure if better represents flow or not.
```
async fn process_header(
	api: &OnlineClient<AvailConfig>,
	header: DaHeader,
	signer: &PairSigner<AvailConfig, Pair>,
) -> Result<()> {
	println!("Finalized block header: {:?}", &header);
	println!("Header data root: {:?}", header.extrinsics_root.data_root);

	let bridge_router_eth_addr: H160 = DA_BRIDGE_ROUTER_ADDRESS.parse()?;
	let tx = avail::tx().da_bridge().try_enqueue_data_root(
		DESTINATION_DOMAIN,
		bridge_router_eth_addr.into(),
		header.into(),
	);

	println!(
		"Sending finalized block header. Domain: {}. Recipient: {}",
		DESTINATION_DOMAIN, DA_BRIDGE_ROUTER_ADDRESS
	);
	let tx_progress = api
		.tx()
		.sign_and_submit_then_watch(&tx, signer, Default::default())
		.await?;

	let h = tx_progress.wait_for_finalized_success().await?;
	let submitted_block = api
		.rpc()
		.block(Some(h.block_hash()))
		.await?
		.ok_or_else(|| anyhow!("no block"))?;

	let xts = submitted_block.block.extrinsics;
	println!("Submitted block extrinsic: {xts:?}");
	Ok(())
}

#[async_std::main]
async fn main() -> Result<(), Box<dyn std::error::Error>> {
	let api = OnlineClient::<AvailConfig>::from_url("ws://127.0.0.1:9944").await?;
	let signer = PairSigner::new(AccountKeyring::Alice.pair());

	let mut finalized_blocks = api.rpc().subscribe_finalized_blocks().await?;

	while let Some(finalized_block) = finalized_blocks.next().await {
		match finalized_block {
			Ok(header) => {
				if let Err(error) = process_header(&api, header, &signer).await {
					println!("Error: {}", error);
				}
			},
			Err(error) => println!("Error: {}", error),
		}

		println!("Sleeping for 10 sec...\n");
		thread::sleep(time::Duration::from_secs(10));
	}

	Ok(())
}
```

Cargo.toml
luketchang: Typo but `//` still gets counted as single slash

pallets/nomad/da-bridge/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+	{
+		/// Dispatch a data root message to the home if the header is valid.
+		#[pallet::weight(100)]
+		pub fn try_dispatch_data_root(
+			origin: OriginFor<T>,
+			destination_domain: u32,
+			recipient_address: H256,
+			header: T::Header,
+		) -> DispatchResult {
+			ensure_signed(origin)?;
```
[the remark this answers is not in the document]
luketchang: Yes, as long as the header was a real block header that the chain finalized, the pallet will accept it. Actor does not need to be trusted because we check header validity

pallets/nomad/da-bridge/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+		H256: From<T::Hash>,
+		u32: From<T::BlockNumber>,
+	{
+		/// Dispatch a data root message to the home if the header is valid.
+		#[pallet::weight(100)]
+		pub fn try_dispatch_data_root(
+			origin: OriginFor<T>,
+			destination_domain: u32,
+			recipient_address: H256,
+			header: T::Header,
```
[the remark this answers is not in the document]
luketchang: Can make this change for `do_dispatch_data_root`. But we do need the full header for `try_dispatch_data_root` because we need to validate the full header by checking block hash is in finalized mapping

pallets/nomad/da-bridge/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+
+	#[pallet::genesis_build]
+	impl<T: Config> GenesisBuild<T> for GenesisConfig {
+		fn build(&self) {}
+	}
+
+	// Hooks
+	#[pallet::hooks]
+	impl<T: Config> Hooks<T::BlockNumber> for Pallet<T> {
+		fn on_finalize(block_number: T::BlockNumber) {
```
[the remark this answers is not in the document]
luketchang: Is the block number given in the on_finalize hook _minus 1_ not guaranteed to be finalized?

pallets/nomad/da-bridge/src/lib.rs
luketchang: How would you suggest clearing the mapping then in bounded way? We run into the same problem in the `home` pallet when clearing the root <> index mappings

pallets/nomad/da-bridge/src/message.rs
luketchang: The receiving solidity contracts expect a 4-byte u32 for block number. We can have this field be generic but ultimately the Solidity contract needs to match the block number size so thought it would make sense to just have it be concrete. Thoughts?

pallets/nomad/da-bridge/src/lib.rs
luketchang: Would you prefer to just have loop have bounded number of iterations? And how to parametrize that?

pallets/nomad/da-bridge/src/lib.rs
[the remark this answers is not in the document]
MiguelDD1: No, it's not. Only GRANDPA can guarantee the finalization. I can see 2 possible solutions here:
- Use an offchain worker, who is going to follow the chain finalization, or
- Integrate your pallet with the finality-grandpa gadget (I need to double-check if this is feasible in our time restriction)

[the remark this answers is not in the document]
luketchang: Final consensus from call is that we allow unfinalized block headers to be dispatched into home's merkle tree BUT we rely on fact that updater will only attest to final data. If the updater doesn't, it will be slashed for submitting an improper update (update that doesn't match actual on-chain state)

pallets/nomad/home/src/lib.rs
```diff
@@ -0,0 +1,403 @@
...
+			Self::do_dispatch(
+				sender.into(),
+				destination_domain,
+				recipient_address,
+				message_body,
+			)
+		}
+
+		/// Verify/submit signed update.
+		#[pallet::weight(100)]
```
MiguelDD1, line 173:
```suggestion
		#[pallet::weight(T::WeightInfo::update())]
```

MiguelDD1 (approved):
Minor change required, we could do it later.

PR #87, opened 2022-09-21
Sync main with develop

aterentic-ethernal:


MiguelDD1 (commented):
IMHO we need a couple of things before sync main:
- **Bump the runtime** version due to header changes. Most likelihood the explorer won't be backward-compatible but at least nodes could synchronize from genesis (hopefully).
-  **Bump the node crate version**, due to new RPC fn.
-  It seems migration code is **not** needed here.

aterentic-ethernal:
> IMHO we need a couple of things before sync main:
> 
>     * **Bump the runtime** version due to header changes. Most likelihood the explorer won't be backward-compatible but at least nodes could synchronize from genesis (hopefully).
> 
>     * **Bump the node crate version**, due to new RPC fn.
> 
>     * It seems migration code is **not** needed here.

So runtime version 5 and 1.1.0 (or 2.0.0)?

MiguelDD1 (approved):

PR #92, opened 2022-09-30
compress amd64 binary, remove centos7

rebelArtists:
**purpose:** 
compress amd64 binaries to match aarch64

**follow-up**
remove uncompressed amd64 binary once downstream ansible playbooks refactored to unarchive compressed version

**tests:**
smoke tests, successful github actions release run: https://github.com/maticnetwork/avail/actions/runs/3161341290

### DIFF
--- a/.github/workflows/releaser.yml
+++ b/.github/workflows/releaser.yml
@@ -8,30 +8,6 @@ on:
 
 jobs:
 
-  # awx compatible binary with correct version of gcc and clang
-  binary_centos7_amd64:
-    runs-on: ubuntu-latest
-    container:
-      image: amd64/centos:centos7
-    steps:
-         - uses: actions/checkout@v2
-         - name: install cargo deps and build avail
-           run: |
-            yum install centos-release-scl -y
-            yum install llvm-toolset-7 -y
-            yum install gcc-c++ -y
-            scl enable llvm-toolset-7 bash
-            export PATH="/opt/rh/llvm-toolset-7/root/usr/bin:$PATH"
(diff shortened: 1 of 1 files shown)


.github/workflows/releaser.yml
```diff
@@ -8,31 +8,23 @@ on:
 
 jobs:
 
-  # awx compatible binary with correct version of gcc and clang
-  binary_centos7_amd64:
```
rebelArtists, line 12: drop awx centos7 builds as no longer required (polycli handles key creation now)

.github/workflows/releaser.yml
```diff
@@ -8,31 +8,23 @@ on:
...
             cargo build --release -p data-avail
-            mv target/release/data-avail target/release/data-avail-centos7-amd64
+            mv target/release/data-avail target/release/data-avail-linux-amd64
          - uses: actions/upload-artifact@v2
            with:
-             name: data-avail-centos7-amd64-binary
-             path: target/release/data-avail-centos7-amd64
-
-  binary_linux_amd64:
+             name: data-avail-linux-amd64-binary
```
rebelArtists, line 24: keep uncompressed amd64 version for now to prevent any downstream breakage, we will delete once ansible playbooks updated to unarchive tar version

.github/workflows/releaser.yml
praetoriansentry: This is great. Simplifies things

praetoriansentry (approved):
lgtm

PR #94, opened 2022-10-06
Miguel/ava 141 hard fork on plonk update

MiguelDD1:


.github/workflows/default.yml
```diff
@@ -28,18 +28,28 @@ jobs:
...
             # After merge, this section should be removed.
             toolchain: nightly-2021-12-05
-            components: rustfmt, clippy, llvm-tools-preview
+            components: rustfmt
+            profile: minimal
+
+      - name: Install Rust toolchain 1.64 
+        uses: actions-rs/toolchain@v1
+        with:
+            toolchain: 1.64
```
aterentic-ethernal, line 44: Any reason for having different toolchain for clippy and llvm-tools? To have more stardnard `-C instrument-coverage`?

node/Cargo.toml
```diff
@@ -18,12 +18,16 @@ name = "data-avail"
 
 [dependencies]
 # Internals
-da-primitives= { path = "../primitives/avail" }
+da-primitives = { git = "https://github.com/maticnetwork/avail-core.git", tag = "kate/v0.2.1", default-features = false }
```
aterentic-ethernal, line 21: Maybe idea for the future, it would be great if we could split primitevies to substrate related ones, (like block header), and avail related once, like matrix data index. Former would be in avail repo, while latter in avail-core, ready for sharing and testing. WDYT?

aterentic-ethernal (commented):
Why are nomad changes ended up in this PR?

.github/workflows/default.yml
MiguelDD1: There is a bug on `cargo` at `nightly-2021-12-05`, so I bumped to recent toolchain `1.64`. 
However I kept the `rust-fmt` to avoid changes on code-style. It seems that `rust-fmt` from `1.64` does not support some features inside `.rustfmt.toml`.

This is the error I got using `nightly-2021-12-05`:

```
❯ cargo b -p data-avail
Updating crates.io index
error: failed to select a version for the requirement `parity-scale-codec = "^3.0.0"`
candidate versions found which didn't match: 3.1.5, 3.1.3, 3.1.2, ...
location searched: crates.io index
required by package `kate v0.2.1 (https://github.com/maticnetwork/avail-core.git?tag=kate/v0.2.1#17c8a30a)`
... which satisfies git dependency `kate` (locked to 0.2.1) of package `fuzzing v0.1.0 (/mnt/storage/home/miguel/polygon/avail_3/fuzzing)`
```

node/Cargo.toml
MiguelDD1: Good idea. I've created a [new ticket](https://linear.app/matic/issue/AVA-396/split-substrate-primitives-and-avail-related-types-on-avail-core) for that.

MiguelDD1:
Superseeded by [PR#96](https://github.com/maticnetwork/avail/pull/96)

MiguelDD1:
Super-seeded by https://github.com/maticnetwork/avail/pull/96

PR #98, opened 2022-10-19
da-bridge actor script takes in env vars

luketchang:
- Modifies `da-bridge` actor subxt script to take environment variables for rpc, tx signer key, interval blocks, destination domain, and da-bridge router address
- User fills out .env file and runs `env $(cat .env | xargs) cargo run --example da_bridge_actor`


MiguelDD1:
@luketchang this needs a rebase

PR #99, opened 2022-10-20
Header Forward Compatibility

MiguelDD1:
# Description
`Header` has been split into 2 parts:
 - A _common header_ which is similar to the generic substrate header,
 - And a _header extension_ field which supports **forward compatibility** using versions.
  
# Other changes
- Data Root build uses `beefy_merkle_tree` and supports `no-std`.

### DIFF
--- a/pallets/system/Cargo.toml
+++ b/pallets/system/Cargo.toml
@@ -13,7 +13,7 @@ readme = "README.md"
 targets = ["x86_64-unknown-linux-gnu"]
 
 [dependencies]
-da-primitives = { git = "https://github.com/maticnetwork/avail-core.git", branch = "miguel/ava-397-compatibility-issue-between-10-11", default-features = false, features = ["header-backward-compatibility-test"] }
+da-primitives = { git = "https://github.com/maticnetwork/avail-core.git", branch = "miguel/ava-397-compatibility-issue-between-10-11", default-features = false }
 kate = { git = "https://github.com/maticnetwork/avail-core.git", branch = "miguel/ava-397-compatibility-issue-between-10-11", default-features = false }
 kate_v1 = { git = "https://github.com/maticnetwork/avail-core.git", tag = "kate/v0.1.0", package = "kate", default-features = false }
 da-primitives_v1 = { git = "https://github.com/maticnetwork/avail-core.git", tag = "kate/v0.1.0", package = "da-primitives", default-features = false }
--- a/pallets/system/Cargo.toml
+++ b/pallets/system/Cargo.toml
@@ -13,16 +13,13 @@ readme = "README.md"
 targets = ["x86_64-unknown-linux-gnu"]
 
 [dependencies]
-da-primitives = { git = "https://github.com/maticnetwork/avail-core.git", branch = "miguel/ava-397-compatibility-issue-between-10-11", default-features = false }
+da-primitives = { git = "https://github.com/maticnetwork/avail-core.git", branch = "miguel/ava-397-compatibility-issue-between-10-11", default-features = false, features = ["header-backward-compatibility-test"]}
 kate = { git = "https://github.com/maticnetwork/avail-core.git", branch = "miguel/ava-397-compatibility-issue-between-10-11", default-features = false }
-kate_v1 = { git = "https://github.com/maticnetwork/avail-core.git", tag = "kate/v0.1.0", package = "kate", default-features = false }
-da-primitives_v1 = { git = "https://github.com/maticnetwork/avail-core.git", tag = "kate/v0.1.0", package = "da-primitives", default-features = false }
-
 # Other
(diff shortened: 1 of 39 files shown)


pallets/system/src/data_root_builder.rs
```diff
@@ -0,0 +1,355 @@
...
+			target: LOG_TARGET,
+			"Used {} extrinsics of {}",
+			used_extrinsics_count,
+			total_extrinsics_count
+		);
+
+		root.into()
+	}
+
+	#[cfg(not(feature = "force-rs-merkle"))]
```
prabal-banerjee, line 81: do you think using two libraries can create inconsistency in block gen / syncing later? A bug in one codebase will lead to creating incompatible headers, right?

pallets/system/src/data_root_builder.rs
```diff
@@ -0,0 +1,355 @@
...
+	}
+
+	fn encoded_submit_call<A: Into<AppId>>(app_id: A) -> AppExtrinsic {
+		let data = hex!("5D0284001CBD2D43530A44705AD088AF313E18F80B53EF16B36177CD4B77B846F2A5F07C01C44755794EA949E9410390CB4CE07FE2D8068656185B5AB9B43EEF934C3680478968C1F83E360A5D942FE75E9D58E49106A8E8B23601CBC6A633D80E5D089D83A4000400030000001D01A46868616A6B616E636B61206C61682069616B6A206361697568206162206169616A6820612067616861").to_vec();
+		AppExtrinsic {
+			app_id: app_id.into(),
+			data,
+		}
+	}
+
```
prabal-banerjee, line 348: do we have test cases checking root parity between the two libs? also, these tests look to be mainly focussing on correct proof of inclusion. do we have checks about correctness of data root generation?

pallets/system/Cargo.toml
```diff
@@ -13,16 +13,13 @@ readme = "README.md"
...
-da-primitives_v1 = { git = "https://github.com/maticnetwork/avail-core.git", tag = "kate/v0.1.0", package = "da-primitives", default-features = false }
-
+da-primitives = { git = "https://github.com/maticnetwork/avail-core.git", tag = "v0.2.0", default-features = false, features = ["header-backward-compatibility-test"]}
+kate = { git = "https://github.com/maticnetwork/avail-core.git", tag = "v0.2.0", default-features = false }
 # Other
 impl-trait-for-tuples = "0.2.1"
 static_assertions = "1.1.0"
 log = { version = "0.4.14", default-features = false }
-rs_merkle = { version = "1.2.0", default-features = false }
+rs_merkle = { version = "1.2.0", default-features = false, optional = true }
```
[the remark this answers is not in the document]
MiguelDD1: I just keep it for testing purpose: there is a test to verify that `beefy_merkle` output is the same as `rs_merkle`. 
We should keep it until the first E2E test case on nomad bridge is done.

pallets/system/src/data_root_builder.rs
MiguelDD1: Actually runtime (in wasm) does NOT compile with `force-rs-merkle` feature. 
This is only to testing the replacement of the library.

pallets/system/src/data_root_builder.rs
MiguelDD1: Yes, it is in `runtime/src/data_root_tests.rs:67`. The check generates the same expected output for both libraries. In order to verify `rs-merkle`, you have to enable the feature like:
```Bash
$> cargo test --workspace --features "frame-system/force-rs-merkle"
```

**Note** that any  Data Root test has been moved to `runtime` level because there is a dependency between the `Call::DataAvailility(da_ctr::Call::submit_data{...})` and the implementation of `DataRootFilter`. That will **make code upgradable** and remove lot of duplicate/hardcode deserialization code.

pallets/system/src/data_root_builder.rs
```diff
@@ -0,0 +1,355 @@
...
+			.filter_map(|(idx, app_extrinsic)| {
+				// Decode call and log any failure.
+				total_extrinsics_count += 1;
+				match F::UncheckedExtrinsic::decode(&mut app_extrinsic.data.as_slice()) {
+					Ok(app_unchecked_extrinsic) => {
+						// Some((app_unchecked_extrinsic, app_extrinsic.data.as_slice()))
+						Some(app_unchecked_extrinsic)
+					},
+					Err(err) => {
+						log::error!(
```
aterentic-ethernal, line 39: This will log error and skip invalid extrinsic? Should case of badly encoded extrinsic stop node and demand bug fix, or its ok to ignore it (and monitor logs)?

prabal-banerjee (approved):

kroos47 (approved):

PR #102, opened 2022-10-26
Miguel/ava 405 rpc not working with scripts

MiguelDD1:
# Description
The `query_data_proof` RPC now uses the new way to filter, extract, and decode the `data` field from `DataAvailability::submit_data` calls (`submitted_call::call_proof`), and it is aligned with the code used inside the runtime, removing duplications.


rpc/kate-rpc/src/lib.rs
```diff
@@ -7,39 +7,41 @@ use codec::{Compact, Decode, Encode, Error as DecodeError, Input};
...
-	) -> Result<Vec<u8>>;
+	fn query_proof(&self, cells: Vec<Cell>, at: Option<HashOf<Block>>) -> Result<Vec<u8>>;
 
 	#[rpc(name = "kate_blockLength")]
-	fn query_block_length(&self) -> Result<BlockLength>;
+	fn query_block_length(&self, at: Option<HashOf<Block>>) -> Result<BlockLength>;
 
 	#[rpc(name = "kate_queryDataProof")]
-	fn query_data_proof(&self, block_number: NumberOrHex, index: u32) -> Result<Vec<H256>>;
+	fn query_data_proof(&self, data_index: u32, at: Option<HashOf<Block>>) -> Result<DataProof>;
```
prabal-banerjee, line 44: does this mean the API now accepts both block number or hash instead of block number earlier?

prabal-banerjee (approved):

rpc/kate-rpc/src/lib.rs
MiguelDD1: No, it only accepts the `hash`. It is a kind of de-facto standard in RPC calls on substrate.

PR #113, opened 2022-11-09
Switching to Nightly Toolchain

praetoriansentry:
The current build process was failing and required use of the nightly toolchain. I'm switching to the nightly tool chain in the CI flow and also committing some standard Rust CI workflows from @cvhpoly.

There is  a successful run here: https://github.com/maticnetwork/avail/actions/runs/3431455773

That run created this release: https://github.com/maticnetwork/avail/releases/tag/v1.3.0-2-ad405d0


jakubcech (approved):

MiguelDD1:
If we use an specific `nightly` version instead of the latest one, we will get **reproducible builds**.

PR #120, opened 2022-11-25
Switch to newer avail-core.

luka-ethernal:


Cargo.lock
```diff
@@ -1426,8 +1426,8 @@ dependencies = [
 
 [[package]]
 name = "da-primitives"
-version = "0.2.2"
-source = "git+https://github.com/maticnetwork/avail-core.git?tag=v0.3.0#0025bb53b4190d5920ad351253ed75e078c3c833"
+version = "0.4.0"
```
MiguelDD1, line 1429: Should not `da-primitives` be also `0.5.0` following the tag?

MiguelDD1 (changes requested):
Please merge into `develop` branch.

pallets/system/src/limits.rs
```diff
@@ -155,8 +155,8 @@ impl BlockLength {
 		const_assert!(is_chunk_size_valid(BLOCK_CHUNK_SIZE));
 		Self {
 			max: PerDispatchClass::new(|_| max),
-			cols: MAX_BLOCK_COLUMNS,
-			rows: MAX_BLOCK_COLUMNS,
+			cols: MAX_BLOCK_COLUMNS.0,
+			rows: MAX_BLOCK_ROWS.0,
```
aterentic-ethernal, line 159: This was a bug?

luka-ethernal: I guess so. I changed it along the way.

PR #127, opened 2022-12-12
testnet v3 chainspec update

gatsbyz:


MiguelDD1 (changes requested):
Please add also the plain Json file.

gatsbyz:
@MiguelDD1 The plain json file is added [here](https://github.com/maticnetwork/avail/blob/main/misc/genesis/testnet-v3.chain.spec.json). This PR updates the raw file because of inconsistencies with the original file.

MiguelDD1 (approved):

MiguelDD1:
As we talked off-line, that file was generated from `https://testnet.polygonavail.net/chaininfo.txt` **and** using the binary tagged as `v1.3.0-2-ad405d0`.

PR #133, opened 2023-01-17
Add kate rows RPC

aterentic-ethernal:
This PR adds `queryRows` RPC to support fetching and pushing rows to DHT on light clients. Since new RPC is added, node version is increased from 1.4.0 to 1.5.0. Dependency to avail-core will be updated once corresponding PR is merged.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,9 +25,9 @@ frame-system-rpc-runtime-api = { path = "pallets/system/rpc/runtime-api" }
 frame-system-benchmarking = { path = "pallets/system/benchmarking" }
 
 # Polygon Primitives
-da-primitives = { git="https://github.com/maticnetwork/avail-core", version = "0.4.2", tag = "da-primitives/v0.4.2" }
-kate = { git="https://github.com/maticnetwork/avail-core", version = "0.5.3", tag = "da-primitives/v0.4.2" }
-kate-recovery = { git="https://github.com/maticnetwork/avail-core", version = "0.7.1", tag = "da-primitives/v0.4.2" }
+da-primitives = { git="https://github.com/maticnetwork/avail-core", version = "0.4.2", tag = "kate-recovery/v0.8.0" }
+kate = { git="https://github.com/maticnetwork/avail-core", version = "0.6.0", tag = "kate-recovery/v0.8.0" }
+kate-recovery = { git="https://github.com/maticnetwork/avail-core", version = "0.8.0", tag = "kate-recovery/v0.8.0" }
 
 
 # Nomad
--- a/rpc/kate-rpc/src/lib.rs
+++ b/rpc/kate-rpc/src/lib.rs
@@ -160,18 +160,19 @@ where
 			.block(&block_id)
 			.map_err(|e| internal_err!("Invalid block number: {:?}", e))?
 			.ok_or_else(|| internal_err!("Missing block {}", block_id))?;
+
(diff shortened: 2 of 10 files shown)


Cargo.toml
```diff
@@ -25,9 +25,9 @@ frame-system-rpc-runtime-api = { path = "pallets/system/rpc/runtime-api" }
 frame-system-benchmarking = { path = "pallets/system/benchmarking" }
 
 # Polygon Primitives
-da-primitives = { git="https://github.com/maticnetwork/avail-core", version = "0.4.2", tag = "da-primitives/v0.4.2" }
-kate = { git="https://github.com/maticnetwork/avail-core", version = "0.5.3", tag = "da-primitives/v0.4.2" }
-kate-recovery = { git="https://github.com/maticnetwork/avail-core", version = "0.7.1", tag = "da-primitives/v0.4.2" }
+da-primitives = { git="https://github.com/maticnetwork/avail-core", version = "0.4.2", branch = "AV-120-rows-in-dht" }
```
MiguelDD1, line 28: Deploy and tag `avail-core` changes before merge this.

rpc/kate-rpc/src/lib.rs
```diff
@@ -141,6 +148,77 @@ where
...
+				block_length.chunk_size(),
+				&xts_by_id,
+				seed,
+			)
+			.map_err(|e| internal_err!("Flatten and pad block failed: {:?}", e))?;
+
+			let data = kate::com::par_extend_data_matrix(block_dims, &block)
+				.map_err(|e| internal_err!("Matrix cannot be extended: {:?}", e))?;
+
+			block_ext_cache.put(block_hash, (data, block_dims));
```
MiguelDD1, line 207: QQ: Can we insert block in cache if and only if that block is finalized?

rpc/kate-rpc/src/lib.rs
```diff
@@ -141,6 +148,77 @@ where
...
+			.map_err(|e| internal_err!("Invalid block number: {:?}", e))?
+			.ok_or_else(|| internal_err!("Missing block {}", block_id))?;
+		let block_hash = signed_block.block.header().hash();
+
+		let mut block_ext_cache = self
+			.block_ext_cache
+			.write()
+			.map_err(|_| internal_err!("Block cache lock is poisoned .qed"))?;
+
+		let block_length: BlockLength = self
```
MiguelDD1, line 170: `block_length` only used inside `if`-block. Pls move it.

Cargo.toml
aterentic-ethernal: Sure, I have mentioned that in description, but didn't want to postpone review.

rpc/kate-rpc/src/lib.rs
aterentic-ethernal: This affects three RPCs, I didn't wanted to spend much there, but RPCs are almost the same and code should be reused at some point.

rpc/kate-rpc/src/lib.rs
aterentic-ethernal: How to check if block is finalized?

Cargo.toml
aterentic-ethernal: Done, but its not possible to use different tags for different crates from same repository, I don't know whats going on but we should revisit this at some point.

MiguelDD1 (approved):

PR #143, opened 2023-03-22
Node binary does not support Runtime Upgrades

fmiguelgarcia:
Use `OpaqueExtrinsic` at node binary level. That makes the binary completely agnostic to the underlying runtime.

## Companion PRs
- [Avail-Core#28](https://github.com/availproject/avail-core/pull/28)
- [Avail-apps#23](https://github.com/availproject/avail-apps/pull/23)

## Side-Effects
- RPC interfaces need to transform from `OpaqueExtrinsic` into `AppUncheckedExtrinsic` manually, like https://github.com/availproject/avail/blob/381e1539e0829b4c3d6551437c88fc9af32ebd22/rpc/kate-rpc/src/lib.rs#L164

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1572,6 +1572,7 @@ dependencies = [
  "frame-support",
  "frame-system",
  "kate",
+ "log",
  "pallet-balances",
  "pallet-transaction-payment",
  "parity-scale-codec 3.4.0",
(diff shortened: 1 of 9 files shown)


0xSasaPrsic (approved):
Nice one!

aterentic-ethernal (approved):

PR #151, opened 2023-04-25
Add a tool/example that verifies grandpa finality of the blocks (chai…

luka-ethernal:
…ns).

### DIFF
--- a/avail-subxt/examples/check_finality.rs
+++ b/avail-subxt/examples/check_finality.rs
@@ -1,0 +1,267 @@
+use avail_subxt::{
+	api::{self},
+	build_client,
+	primitives::Header,
+};
+use codec::{Decode, Encode};
+use serde::{de::Error, Deserialize};
+use sp_core::{
+	blake2_256, bytes,
+	ed25519::{self, Public as EdPublic, Signature},
+	Pair, H256,
+};
+use subxt::rpc_params;
+use tokio::sync::mpsc::unbounded_channel;
+
+#[derive(Debug, Encode)]
+pub enum SignerMessage {
+	DummyMessage(u32),
+	PrecommitMessage(Precommit),
+}
+
+#[derive(Clone, Debug, Decode, Encode, Deserialize)]
+pub struct Precommit {
+	pub target_hash: H256,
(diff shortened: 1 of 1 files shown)


aterentic-ethernal (approved):

kroos47 (approved):

avail-subxt/examples/check_finality.rs
```diff
@@ -0,0 +1,267 @@
...
+
+	// Forming a channel for sending any relevant events gathered asynchronously through Substrate WS API.
+	let (msg_sender, mut msg_receiver) = unbounded_channel::<Messages>();
+
+	// Task that produces headers and new validator sets
+	tokio::spawn({
+		let subxt_client = subxt_client.clone();
+		let msg_sender = msg_sender.clone();
+		async move {
+			while let Some(Ok(header)) = header_subscription.next().await {
```
prabal-banerjee, line 126: Does this also suffer the same problem of jumping blocks in case of delayed finality? The `subxt` examples also do it like this, so I'm guessing it gets handled internally?

prabal-banerjee: Although, on second thought, this particular case is an exception. Auth set change containing blocks are always finalized. But still my original question remains for my general understanding.

prabal-banerjee (approved):

avail-subxt/examples/check_finality.rs
luka-ethernal: Yes, this also suffers the same issue of skipping blocks. But for verifying finality and changes in auth set, it's not a problem, since changing authset is always in a block signed by supermajorty, so it's always properly verified. And I did test it too. I paused finality until the change of session/new authset and resumed it after and I received a header containing changes in auth set and resumed normally.

In this example I don't backtrack the headers till the last checkpoint, because there's no need to. If the supermajority agrees the block is the new head, that's enough. For sampling and appdata purposes, this will have to be done, of course.

prabal-banerjee: Got it. Makes sense. Thanks!

PR #152, opened 2023-05-02
Rename scripts

kroos47:


tests/gsrpc/README.md
```diff
@@ -11,8 +11,9 @@ PS: Do keep in mind that the gsprc is not that stable like polakdot-js and subxt
     "seed":"bottom drive obey lake curtain smoke basket hold race lonely fit walk//Alice",
     "api_url":"ws://127.0.0.1:9944",
     "size":1000,
-    "app_id" : 0
-
+    "app_id" : 0,
+    "dest" : "0x8eaf04151687736326c9fea17e25fc5287613693c912909cb226aa4794f26a48",
```
prabal-banerjee, line 15: Can't we make it account ID instead of hex?

tests/gsrpc/README.md
```diff
@@ -45,6 +46,15 @@ go run extrinsics/dataSubmitAndWatch/dataSubmitAndWatch.go -config config.json
...
 The same procedure of data submission is done here. But checks/logs if the data is included in the block. The default APP_ID is 0
 
+>Balance Transfer
+
+```
+go run extrinsics/balance/balance.go -config config.json
+```
+
+Amount is transferred from the seed account to the dest account with the amount mentioned in config file. Note that the dest account mentioned must be using the public hex key. 
+You can refer [here](https://docs.substrate.io/tutorials/build-a-blockchain/add-trusted-nodes/#generate-your-account-and-keys) on how to create or inspect account
```
prabal-banerjee, line 56: Can we account ID instead? By using this function instead of hex:
https://pkg.go.dev/github.com/centrifuge/go-substrate-rpc-client@v2.0.0+incompatible/types#NewAddressFromAccountID

tests/js_api/src/transfer.ts
```diff
@@ -56,7 +61,13 @@ async function Transfer(api: ApiPromise) {
...
     const api = await createApi();
     const metadata = await api.rpc.state.getMetadata();
-    console.log(config.amount);
+    const [chain, nodeName, nodeVersion] = await Promise.all([
+        api.rpc.system.chain(),
+        api.rpc.system.name(),
+        api.rpc.system.version()
+      ]);
+    
+      console.log(`You are connected to chain ${chain} using ${nodeName} v${nodeVersion}`);
```
prabal-banerjee, line 70: nit: not sure if indentation is correct.

prabal-banerjee (approved):
Can we add some tests to the CI/CD pipeline to make sure that these pass for every new version? Maybe add it to https://github.com/availproject/avail/blob/main/.github/workflows/default.yml ?

prabal-banerjee (approved):

tests/gsrpc/README.md
kroos47: yeh converted the account address to hexstring directly then used the same function

PR #154, opened 2023-05-16
Add DA Protocol during the Block Import process

fmiguelgarcia:
It adds a new worker to the import process of the block, where we double-check our header extensions (i.e. `kate commitment` and `data root`.


Aphoh (approved):
At a glance looks good. Plugging in the header checks in the block import process is definitely the right place to put it. 

@fmiguelgarcia did you manage to find any other places the headers were being built? IIRC they were still getting built on import but we couldn't find exactly where the builder was getting called.

pallets/system/src/header_builder.rs
```diff
@@ -62,21 +62,22 @@ pub trait HeaderExtensionBuilder {
...
-		let seed = <T as Config>::Hashing::hash_of(&(&epoch_seed, &block_number));
-
-		log::trace!(
-			target: LOG_TARGET,
-			"Header builder seed {:?} from epoch seed {:?} and block {:?}",
-			seed,
-			epoch_seed,
-			block_number
-		);
+		let seed = if cfg!(feature = "secure_padding_fill") {
```
Leouarz, line 71: The goal here is to run the testnet without this feature so there's not epoch issue ?

node/src/da_block_import.rs
```diff
@@ -0,0 +1,116 @@
...
+			);
+
+		self.inner
+			.import_block(block, new_cache)
+			.await
+			.map_err(Into::into)
+	}
+
+	/// # TODO
+	/// - Check that `Runtime::System::BlockLenght` was not changed inside the block;
```
fmiguelgarcia, line 109: See https://github.com/availproject/engineering/issues/51

pallets/system/src/header_builder.rs
fmiguelgarcia, line 71: See https://github.com/availproject/engineering/issues/52 to get more info about `secure_padding_fill`

Leouarz: Thanks !

rpc/kate-rpc/src/lib.rs
```diff
@@ -238,23 +234,21 @@ where
...
+			let block_length: BlockLength = self
 				.client
 				.runtime_api()
-				.get_babe_vrf(&block_id)
-				.map_err(|e| internal_err!("Babe VRF not found for block {}: {:?}", block_id, e))?;
+				.block_length(&block_id)
+				.map_err(|e| internal_err!("Block Length cannot be fetched: {:?}", e))?;
+
+			// Use Babe's VRF
+			let seed: [u8; 32] =
```
fmiguelgarcia, line 248: :lady_beetle:  See https://github.com/availproject/engineering/issues/53

rpc/kate-rpc/src/lib.rs
```diff
@@ -329,23 +323,21 @@ where
...
```
fmiguelgarcia, line 337: :lady_beetle: See https://github.com/availproject/engineering/issues/53

fmiguelgarcia:
> At a glance looks good. Plugging in the header checks in the block import process is definitely the right place to put it.
> 
> @fmiguelgarcia did you manage to find any other places the headers were being built? IIRC they were still getting built on import but we couldn't find exactly where the builder was getting called.

Basically I've found 2 places where the builder is called:
- At runtime level, on `system::finalize()`.
- At node level, during the block import process.

pallets/system/src/header_builder.rs
fmiguelgarcia: Yes, that is.
BTW we've also created a ticket to review the security implications and possible solutions [here](https://github.com/availproject/engineering/issues/52) before mainnet.

PR #168, opened 2023-06-05
add a check on the block weight in case of reducing dimensions

Leouarz:


avail-subxt/examples/submit_block_length_proposal.rs
```diff
@@ -0,0 +1,236 @@
...
+	let batch_call = api::tx().utility().batch(vec![sudo_call]);
+	tx_send!(client, &batch_call, signer)
+		.find_first::<DaEvent::BlockLengthProposalSubmitted>()?
+		.expect("2 - Block Length Proposal Submitted event is emitted .qed");
+	println!("2 - Block Length Proposal Submitted found.");
+	reset(client, signer).await?;
+	Ok(())
+}
+
+// TODO: See if submit_block_length_proposal needs to work with TC, Council and/or democracy
```
Leouarz, line 132: Here, we don't use the tech committee test because in the current state, the `submit_block_length_proposal` is not allowed to process in tech committee / council / democracy / ... (only sudo)

fmiguelgarcia (approved):

kroos47 (approved):

PR #170, opened 2023-06-09
[WIP] Add benchmarks for `dataRoot` & `kate commitment` builds

fmiguelgarcia:
It adds benchmarks for some parts we do at header build, at the very end of the block production. We need to figure out if we can add extra-weight (into `Weight.proof_size` part) before the execution of the extrinsic. That will give us better boundaries on the block build process. 
The sub-elements we need to evaluate are:
- `DataRoot` calculation. I guess that will be related with the size of the data parameter.
- `Kate Commitment`. It could be constrained by the current block dimension (`rows` & `cols`) and by the number of extrinsic.

### DIFF
--- a/pallets/dactr/src/benchmarking.rs
+++ b/pallets/dactr/src/benchmarking.rs
@@ -1,11 +1,32 @@
 #![cfg(feature = "runtime-benchmarks")]
 
+use codec::{Decode, Encode};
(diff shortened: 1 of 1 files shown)


Leouarz:
Mentioned in https://github.com/availproject/avail/pull/171

PR #171, opened 2023-06-14
Ghali/improve benchmarks

Leouarz:
Here we can see the first result of benchmarking `data_root` and `commitment` generation.
For the first one, the result seem ok but for kate it's harder because we have 3 parameters, the number of rows, number of columns and the length of the extrinsics. Also the commitment generation is STD only.

cc. @fmiguelgarcia @kroos47


Leouarz:
You will see a quite a number of test where I try to change some of the parameters, I also get some issues when trying a loop on all the parameters.

Leouarz:
PR Created from this : https://github.com/availproject/avail/pull/170

kroos47 (approved):

pallets/dactr/src/lib.rs
```diff
@@ -122,7 +122,7 @@ pub mod pallet {
 		}
 
 		#[pallet::call_index(1)]
-		#[pallet::weight(T::WeightInfo::submit_data(data.len() as u32))]
+		#[pallet::weight(({T::WeightInfo::submit_data(data.len() as u32) + T::WeightInfo::data_root(data.len() as u32)}, DispatchClass::Normal))]
```
fmiguelgarcia, line 125: Please use `saturated_add` and `saturated_into` instead of `+` or `as u32`.
What do you think about using a helper function like [this](https://github.com/availproject/avail/blob/cea467c57619e7d606a3ccdf9615c280771dcb9b/pallets/system/src/lib.rs#L1877) ?
Something like:
```Rust
#[pallet::weight(weight_helper::submit_data(data.len())]
```

pallets/dactr/src/benchmarking.rs
```diff
@@ -1,23 +1,169 @@
...
+	) -> TransactionValidity {
+		Ok(Default::default())
+	}
+}
+
+fn repeat_bytes(byte: u8, len: u32) -> Vec<u8> {
+	repeat(byte).take(len as usize).collect::<Vec<_>>()
+}
+
+fn prev_power_of_two(n: u32) -> u32 {
```
fmiguelgarcia, line 86: Minor thing: Does it repeat several times the same power of 2? Kind of nested iteration if we thought about `--repeat=10` parameter when we run the benchmark.
Is there any way to fix also the rows to `[128,256,512,1024]` ? I guess benchmarks will be faster and we will see less deviation.

Leouarz: The goal was to actually have meaningful repetitions. We can fix rows when using power of 2s for the range instead of plain rows. In the end I set it up like this for better comprehension, but I can revert it if you prefer. (I think since we define steps and repetitions, end result should be same)

PR #173, opened 2023-06-28
Fix the `set_code`'s weights

fmiguelgarcia:
It fixes the `system::set_code(..)`'s weights, using `max_extrinsic` weight instead of `max_block`.

It also changes 2 things (att @prabal-banerjee ):
- Max Ref Weight for a block has been increased from 2 secs to 5 secs (only in non `fast-runtime`)
- Max Proof Size has been reduced from `u64:MAX` to `500ms`, which we could use in the future to handle `data-root` & `kate-commitment` built.

### DIFF
--- a/runtime/src/constants.rs
+++ b/runtime/src/constants.rs
@@ -100,7 +100,9 @@ pub mod time {
 
 pub mod system {
 	use da_primitives::NORMAL_DISPATCH_RATIO;
-	use frame_support::weights::constants::{ExtrinsicBaseWeight, WEIGHT_REF_TIME_PER_SECOND};
+	use frame_support::weights::constants::{
+		ExtrinsicBaseWeight, WEIGHT_REF_TIME_PER_MILLIS, WEIGHT_REF_TIME_PER_SECOND,
+	};
 	use frame_system::limits::BlockWeights as SystemBlockWeights;
 
 	use super::*;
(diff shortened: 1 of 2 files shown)


runtime/src/constants.rs
```diff
@@ -112,10 +114,21 @@ pub mod system {
 	/// We assume that ~10% of the block weight is consumed by `on_initialize` handlers.
 	/// This is used to limit the maximal weight of a single extrinsic.
 	const AVERAGE_ON_INITIALIZE_RATIO: Perbill = Perbill::from_percent(10);
+
+	/// Proof size allowed up to 500ms.
```
kroos47, line 118: does it affect with performance?

Leouarz:
Was there an issue with using the scheduler ?

fmiguelgarcia:
> Was there an issue with using the scheduler ?

Yes, actually it raises up an `Scheduler::PermanentlyOverweight` event.
att @Leouarz

Leouarz (approved):

0xSasaPrsic (approved):

prabal-banerjee:
> Max Proof Size has been reduced from u64:MAX to 500ms, which we could use in the future to handle data-root & kate-commitment built.

I'm not sure why this is needed. The commitment generation can easily take more than 500ms IMO and what would be the reason to not allow beyond that?

PR #174, opened 2023-06-30
update size per tx

kroos47:
updated the tx size from 16KB to 512KB.


pallets/dactr/src/weights.rs
```diff
@@ -18,20 +18,19 @@
 //! Autogenerated weights for da_control
 //!
 //! THIS FILE WAS AUTO-GENERATED USING THE SUBSTRATE BENCHMARK CLI VERSION 4.0.0-dev
-//! DATE: 2023-06-28, STEPS: `20`, REPEAT: 10, LOW RANGE: `[]`, HIGH RANGE: `[]`
-//! HOSTNAME: `ip-172-31-12-189`, CPU: `Intel(R) Xeon(R) Platinum 8175M CPU @ 2.50GHz`
-//! EXECUTION: Some(Wasm), WASM-EXECUTION: Compiled, CHAIN: Some("dev"), DB CACHE: 1024
+//! DATE: 2023-06-30, STEPS: `50`, REPEAT: 25, LOW RANGE: `[]`, HIGH RANGE: `[]`
+//! HOSTNAME: `ip-172-31-39-104`, CPU: `Intel(R) Xeon(R) Platinum 8375C CPU @ 2.90GHz`
+//! EXECUTION: Some(Native), WASM-EXECUTION: Compiled, CHAIN: Some("dev"), DB CACHE: 1024
```
fmiguelgarcia, line 23: Execution should be using `Wasm` 
BTW the reference machine's CPU has changed, which one should we use?

Leouarz: This machine is way more powerful than the one I was using, is it wanted ?

Leouarz:
There are other places where the size need to be changed (also one test if i'm not mistaken)

Leouarz (approved):

Leouarz:
Should we incvrease spec version ?

prabal-banerjee:
> Should we increase spec version ?

Yes, unless there is some major concern. Also, let's make sure LCs are working with the updated spec version.

0xSasaPrsic (approved):

PR #177, opened 2023-07-05
add support for utility pallet in data root generation

Leouarz:


prabal-banerjee (approved):

pallets/system/src/data_root_builder.rs
0xSasaPrsic, line 1: is this used anywhere?

Leouarz: I don't think so, probably only for testing, i changed only to be thorough

0xSasaPrsic (approved):

runtime/src/lib.rs
```diff
@@ -168,32 +168,43 @@ parameter_types! {
...
 			_ => None,
 		}
 	}
+
+	/// This function processes a list of calls and returns their data as Vec<Vec<u8>>
+	fn process_calls(
+		calls: Vec<RuntimeCall>,
+		metrics: &submitted_data::RcMetrics,
+	) -> Option<Vec<Vec<u8>>> {
+		let mut result = Vec::new();
```
fmiguelgarcia, line 196: nit: Reduce possible reallocations
```suggestion
		let mut result = Vec::with_capacity(calls.len());
```

fmiguelgarcia (approved):

pallets/dactr/src/extensions/check_app_id.rs
```diff
@@ -27,49 +28,82 @@ use crate::{Call as DACall, Config, Pallet};
...
-	T: Config + Send + Sync,
-	T::RuntimeCall: IsSubType<DACall<T>>,
+	T: DAConfig + UtilityConfig + Send + Sync,
+	<T as frame_system::Config>::RuntimeCall:
+		IsSubType<DACall<T>> + IsSubType<UtilityCall<T>> + From<UtilityCall<T>>,
 {
 	/// utility constructor. Used only in client/factory code.
 	pub fn from(app_id: AppId) -> Self { Self(app_id, sp_std::marker::PhantomData) }
 
+	/* 	pub fn validate_batch_call(&self, call: &<T as pallet_utility::Config>::RuntimeCall) {} */
```
fmiguelgarcia, line 45: Remove commented code

pallets/dactr/src/extensions/check_app_id.rs
```diff
@@ -27,49 +28,82 @@ use crate::{Call as DACall, Config, Pallet};
...
+		&self,
+		call: &<T as frame_system::Config>::RuntimeCall,
+		maybe_next_app_id: &mut Option<AppId>,
+	) -> Result<(), TransactionValidityError> {
+		let done = match call.is_sub_type() {
+			Some(UtilityCall::<T>::batch { calls })
+			| Some(UtilityCall::<T>::batch_all { calls })
+			| Some(UtilityCall::<T>::force_batch { calls }) => {
+				for call in calls.iter() {
+					let cast = call.clone().into();
```
fmiguelgarcia, line 61: I am worry about this, because it duplicates `call` in memory in a recursive call, so we could need something like factorial of `max_extrinsic_len`. 
I mean, an attacker could create a nested call like:
```
// Number of nested `batchs` is `(max_extrinsic_len - remark_len) / batch_len`.
batch( batch( batch( ... remark)))
```
An it would consume the memory of the node.

I think there are some options here (like "The Good, The Bad, and the Ugly" :D ):
- a) Keep the recursive version, but avoid the clone in some way (references + lifetimes). Also c)
- b) Refactor it to an iterative solution. Also c)
- c) Just limit the number of recursions (or supported nested callls).

I think that implementing only option `c` is not good enough because of memory footprint could be really high even with a low number of max nested calls.

Leouarz: @fmiguelgarcia I was also wondering, since this is for appIds and they will eventually be controlled by some entities. What would be the side effect of just allowing batches to have any valid appIds ? That would definitely simplify the issue

markopoloparadox: I added a temporary solution where we do an actual cast. So far it's unsafe but I hope that @Aphoh could help me to make it safe.

I checked it with the debugger and everything seemed alright. Executing `p call` and `p c` on line 74 gave me the same output: `(&da_runtime::RuntimeCall) call = 0x00007fce580026b0`

Also I have added a limit to the maximum amount of recursion that can happen.

prabal-banerjee (approved):

pallets/dactr/src/extensions/check_app_id.rs
```diff
@@ -27,49 +32,86 @@ use crate::{Call as DACall, Config, Pallet};
...
+							// TODO @Marko @Ghali Error name - Replace with MaxRecursionExceeded when PR gets merged
+							InvalidTransaction::Custom(
+								InvalidTransactionCustomId::ForbiddenAppId as u8
+							)
+						);
+						for call in calls.iter() {
+							// TODO @Marko Make it readable
+							let a = call as *const <T as UtilityConfig>::RuntimeCall;
+							let b = a as *const <T as SystemConfig>::RuntimeCall;
+							let c = unsafe { &*b };
```
fmiguelgarcia, line 90: I would not use this kind of casting, because it depends on how `RuntimeCall` is defined, and that is generated by the specific version of substrate. I mean, a future change in substrate version could break this.

fmiguelgarcia (changes requested):

pallets/dactr/src/extensions/check_app_id.rs
```diff
@@ -27,49 +32,86 @@ use crate::{Call as DACall, Config, Pallet};
...
-				let next_app_id = <Pallet<T>>::peek_next_application_id();
+	///  - `DataAvailability::submit_data(..)` extrinsic can use `AppId != 0`.
+	///  - `Utility::batch/batch_all/force_batch(..)` extrinsic can use `AppId != 0` If the wrapped calls are ALL `DataAvailability::submit_data(..)`.
+	///  - Any other call must use `AppId == 0`.
+	pub fn do_validate(&self, call: &<T as SystemConfig>::RuntimeCall) -> TransactionValidity {
+		if self.app_id() == AppId(0) {
+			return Ok(ValidTransaction::default());
+		}
+
+		let mut stack = Vec::new();
```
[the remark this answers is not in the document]
Leouarz: @fmiguelgarcia In here the stack is used to track all calls but iteration is incremented only when a call from utility pallet is detected.

pallets/dactr/src/extensions/check_app_id.rs
markopoloparadox: Removed unsafe code. Utility::RuntimeCall implements trait `IsType<<Self as frame_system::Config>::RuntimeCall>` which means that is trivial to convert to and back from `< frame_system::Config>::RuntimeCall`

pallets/dactr/src/extensions/check_app_id.rs
markopoloparadox: Removed unsafe code. Utility::RuntimeCall implements trait IsType<<Self as frame_system::Config>::RuntimeCall> which means that is trivial to convert to and back from < frame_system::Config>::RuntimeCall

kroos47 (approved):

fmiguelgarcia (approved):

markopoloparadox (approved):

markopoloparadox (approved):

PR #178, opened 2023-07-06
fix weight on set_code when using scheduler

Leouarz:


0xSasaPrsic (approved):

luka-ethernal (approved):

kroos47 (approved):

markopoloparadox:
Looks like we changes the same code in two different but similar PRs: https://github.com/availproject/avail/pull/180
Let's have a call and discuss what the next move should be.

PR #182, opened 2023-07-25
Switch to decoding extrinsics inside runtime instead of kate RPC to a…

0xfluxx:
…void problems with runtime upgrades. Don't silently filter out extrinsics. Log issues.

### DIFF
--- a/pallets/system/src/submitted_data.rs
+++ b/pallets/system/src/submitted_data.rs
@@ -90,7 +90,7 @@ where
 }
 
 /// Extracts just the AppId from the extrinsic (if it is signed), otherwise returns None.
-fn extract_app_id_and_inspect<AE>(opaque: &OpaqueExtrinsic) -> Option<AppId>
+pub fn extract_app_id<AE>(opaque: &OpaqueExtrinsic) -> Option<AppId>
 where
 	AE: AppIdExtractor,
 	AE::Error: Debug,
@@ -197,17 +197,6 @@ where
 		.collect::<Vec<_>>();
 
 	proof(submitted_data, data_index, Rc::clone(&metrics))
-}
-
-/// Extracts the AppId from extrinsic, if it is signed, returns None otherwise.
-pub fn extract_app_id<AE>(extrinsic: &OpaqueExtrinsic) -> Option<AppId>
-where
-	AE: AppIdExtractor,
-	AE::Error: Debug,
-{
-	let app_id = extract_app_id_and_inspect::<AE>(extrinsic);
-
-	app_id
 }
 
 /// Creates the Merkle Proof of the submitted data items in `calls` filtered by `F` and
--- a/pallets/system/src/submitted_data.rs
+++ b/pallets/system/src/submitted_data.rs
@@ -1,6 +1,6 @@
 use core::fmt::Debug;
 
-use avail_core::OpaqueExtrinsic;
+use avail_core::{AppId, OpaqueExtrinsic};
 use beefy_merkle_tree::{merkle_proof, merkle_root, verify_proof, Leaf, MerkleProof};
 use sp_core::H256;
 use sp_runtime::traits::Keccak256;
@@ -39,6 +39,22 @@ pub trait Extractor {
 	) -> Result<Vec<Vec<u8>>, Self::Error>;
 }
 
+/// Extracts the `AppId` field from signed extrinsics.
+pub trait AppIdExtractor {
+	type Error: Debug;
+	/// Returns the `AppId` field of `encoded_extrinsic` if it is a signed extrinsic.
+	///
+	/// The `metrics` will be used to write accountability information about the whole process.
+	fn extract(extrinsic: &OpaqueExtrinsic, metrics: RcMetrics) -> Result<AppId, Self::Error>;
+}
+
+#[cfg(any(feature = "std", test))]
+impl AppIdExtractor for () {
+	type Error = ();
+
+	fn extract(_: &OpaqueExtrinsic, _: RcMetrics) -> Result<AppId, ()> { Ok(AppId(0)) }
+}
+
 #[cfg(any(feature = "std", test))]
 impl Extractor for () {
 	type Error = ();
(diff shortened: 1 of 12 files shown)


rpc/kate-rpc/src/lib.rs
```diff
@@ -131,6 +137,24 @@ where
 	}
 }
 
+fn app_id_from_opaque(opaque: &OpaqueExtrinsic) -> Result<AppId, String> {
```
luka-ethernal, line 140: This fn is unused, I just left it (temporarily) to demonstrate how to only extract the AppId, without decoding the call and triggering BoundedVec limit check in native code.

rpc/kate-rpc/src/lib.rs
```diff
@@ -173,8 +197,13 @@ where
 				.block
 				.extrinsics()
 				.iter()
-				.filter_map(|opaque| UncheckedExtrinsic::try_from(opaque).ok())
```
luka-ethernal, line 176: The problem arises here, filter_map shouln't be used here, there is no correct behavior that skips any of the extrinsics. In this instance (the runtime upgrade bug for increasing tx size), the UncheckedExtrinsic::try_from fails because the decoding happens in the native code, which has an old tx size (BoundedVec bound), while the submitted extrinsic with bigger size was submitted and was checked by runtime and accepted into block.

rpc/kate-rpc/src/lib.rs
```diff
@@ -409,16 +448,10 @@ where
 			.ok_or_else(|| internal_err!("Missing block hash {:?}", at))?
 			.block;
 
-		// Get Opaque Extrinsics and transform into AppUncheckedExt.
-		let calls = block
-			.extrinsics()
-			.iter()
-			.filter_map(|opaque| UncheckedExtrinsic::try_from(opaque).ok())
```
luka-ethernal, line 416: This seemed to suffer from the same problem, except here, we need calls to be decoded, app_id is not enough. The decoded extrinsics are then sent to runtime to filter out just Avail Submit Data extrinsics, and the proof is then created (also inside runtime). Upon closer inspection, there is already a public function called extrinsics_proof that instead does the decoding also in runtime, which solves the problem. However, that was probably used to save memory, because in this instance, all the extrinsics are sent over to runtime, instead of just Avail data. There's a todo there about exact length iterator requirements, which makes me think that.

luka-ethernal:
Hey guys, looking for your comments/discussion on this PR, regarding fixing the bug about runtime upgrades. This PR solves the problem, but there are multiple ways to go about this, and for some, I'm not aware of the performance inpact it may have. I've tested the branch to make sure it works, but I didn't do any benchmarking.

pallets/system/src/submitted_data.rs
```diff
@@ -162,6 +191,18 @@ where
 	proof(submitted_data, data_index, Rc::clone(&metrics))
 }
 
+/// Extracts the AppId from extrinsic, if it is signed, returns None otherwise.
+pub fn extract_app_id<AE>(extrinsic: &OpaqueExtrinsic) -> Option<AppId>
```
markopoloparadox, line 195: Can `extract_app_id_and_inspect` and `extract_app_id` be merged into one function?

luka-ethernal: Yes, it's basically just an unnecessary step, but I did it to match other similar functions.

prabal-banerjee:
About this [commit](https://github.com/availproject/avail/commit/0b67a326488fdd695fe65214962d3967c01441c0) `Switch to decoding extrinsics inside runtime instead of kate RPC to avoid problems with runtime upgrades. Don't silently filter out extrinsics. Log issues.` 

Is doing it within the runtime slower than native? Is caching and other features still work the same? Is there any regression test related to this?

rpc/kate-rpc/src/lib.rs
0xSasaPrsic: would be good to se what are benchmark performance on `extrinsics_proof` vs `calls_proof`

rpc/kate-rpc/src/lib.rs
0xSasaPrsic: maybe to add some todo so we can delete it later if it is not used?

rpc/kate-rpc/src/lib.rs
```diff
@@ -246,8 +275,13 @@ where
 				.block
 				.extrinsics()
 				.iter()
-				.filter_map(|opaque| UncheckedExtrinsic::try_from(opaque).ok())
-				.map(AppExtrinsic::from)
+				.map(|e| {
+					let appid = submitted_data::extract_app_id::<Runtime>(e).unwrap_or(AppId(0));
```
fmiguelgarcia, line 279: nit: Could we dedup this from line 201 and 370?

runtime/src/lib.rs
```diff
@@ -213,6 +213,18 @@ impl submitted_data::Extractor for Runtime {
 	}
 }
 
+/// Decodes and extracts the `AppId` of signed extrinsics.
+impl submitted_data::AppIdExtractor for Runtime {
+	type Error = codec::Error;
+
+	fn extract(opaque: &OpaqueExtrinsic) -> Result<AppId, Self::Error> {
+		let extrinsic = UncheckedExtrinsic::try_from(opaque)?;
+		let appid = extrinsic.signature.map(|e| e.2 .8 .0).unwrap_or(AppId(0));
```
fmiguelgarcia, line 222: Type safe:
```suggestion
		let appid = extrinsic.signature.map(|(_a, _s, extra)| extra.app_id()).unwrap_or_default();
```

luka-ethernal: Thanks, will change!

rpc/kate-rpc/src/lib.rs
fmiguelgarcia: Let's delete it as we do not use it.

rpc/kate-rpc/src/lib.rs
```diff
@@ -173,8 +197,13 @@ where
...
```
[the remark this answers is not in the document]
fmiguelgarcia: Hi @luka-ethernal 
Are you sure this is running inside the runtime (WASM part)?. 
I thought the only way to do it is though the `self.client().runtime_api().<function>`, where the executor decides if we can use `native` or `wasm`.

sh3ll3x3c (approved):

aterentic (approved):

aterentic-ethernal (approved):

rpc/kate-rpc/src/lib.rs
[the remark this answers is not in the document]
luka-ethernal: Well, it has been a month, so I'm no longer sure of anything. But what I'm sure is that I've tested the code and it indeed does invoke runtime api at some point, as the extrinsic gets properly decoded. This means we're no longer tripping the BoundedVec<> limit, which means it's using the constant from newest runtime.

PR #188, opened 2023-07-27
add a filter for block origin during block import commitment verifica…

Leouarz:
…tion

Closes: https://github.com/availproject/engineering/issues/137

### DIFF
--- a/node/src/da_block_import.rs
+++ b/node/src/da_block_import.rs
@@ -57,54 +57,49 @@ where
 		block: BlockImportParams<B, Self::Transaction>,
 		new_cache: HashMap<CacheKeyId, Vec<u8>>,
 	) -> Result<ImportResult, Self::Error> {
-		// We only want to check for blocks that are not from "Own" and when the network is fully synced
-		match block.origin {
-			BlockOrigin::NetworkBroadcast | BlockOrigin::ConsensusBroadcast => {
-				let no_extrinsics = vec![];
-				let extrinsics = block.body.as_ref().unwrap_or(&no_extrinsics);
-				let best_hash = self.client.info().best_hash;
-				let block_id = BlockId::Hash(best_hash);
+		// We only want to check for blocks that are not from "Own"
+		if !matches!(block.origin, BlockOrigin::Own) {
+			let no_extrinsics = vec![];
+			let extrinsics = block.body.as_ref().unwrap_or(&no_extrinsics);
+			let best_hash = self.client.info().best_hash;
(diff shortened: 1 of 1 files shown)


node/src/da_block_import.rs
```diff
@@ -57,48 +57,55 @@ where
 		block: BlockImportParams<B, Self::Transaction>,
 		new_cache: HashMap<CacheKeyId, Vec<u8>>,
 	) -> Result<ImportResult, Self::Error> {
-		let no_extrinsics = vec![];
-		let extrinsics = block.body.as_ref().unwrap_or(&no_extrinsics);
-		let best_hash = self.client.info().best_hash;
-		let block_id = BlockId::Hash(best_hash);
+		// We only want to check for blocks that are not from "Own" and when the network is fully synced
+		match block.origin {
+			BlockOrigin::NetworkBroadcast | BlockOrigin::ConsensusBroadcast => {
```
prabal-banerjee, line 62: Bcoz this is so important in terms of security of the chain, can we write some test case? I imagine that might be bit hard. 

What is the diff between `NetworkBroadcast` and `ConsensusBroadcast`?

Leouarz: @prabal-banerjee 
Indeed, automated testing would be quite hard, but i extensively tested it on my end.

In here : https://paritytech.github.io/substrate/master/sp_consensus/enum.BlockOrigin.html

We can see the different type of block origin

We can see that we can safely exclude Genesis, InitialSync cause blocks are already finalized, Own are also unwanted, File means that the blocks are coming from a file DB.

NetworkBroadcast is the most common one when the block was simply broadcasted.
ConsensusBroadcast is a block that was received from the network and validated in the consensus process.

Also in this code: https://github.com/paritytech/substrate/blob/master/client/service/src/client/client.rs#L626
It's quite explicit how the filter works and used in substrate, the only difference we have is that we also exclude Own.
I also tested this code by running a full node without any issue.

prabal-banerjee (approved):

node/src/da_block_import.rs
fmiguelgarcia: It's not an strong option here but I would also add:
-  `NetworkInitialSync` because a fork due to fail DA protocol could be accepted if we use a bug/misbehave node to synch with.
- `File` because it could help on future testing or if we use it directly from cli.

So I'd only sort-circuited the `Own` origin.

fmiguelgarcia (approved):

PR #198, opened 2023-08-17
Add tx index from block mapping to data index.

0xSasaPrsic:
Add mapping of da transaction index from the block to data index in the merkle trie when generating data root proof.

### DIFF
--- a/rpc/kate-rpc/src/lib.rs
+++ b/rpc/kate-rpc/src/lib.rs
@@ -61,7 +61,7 @@ where
 	#[method(name = "kate_queryDataProof")]
 	async fn query_data_proof(
 		&self,
-		data_index: u32,
+		transaction_index: u32,
 		at: Option<HashOf<Block>>,
 	) -> RpcResult<DataProof>;
 }
--- a/rpc/kate-rpc/src/lib.rs
+++ b/rpc/kate-rpc/src/lib.rs
@@ -9,7 +9,7 @@ use avail_core::{
 	header::HeaderExtension, traits::ExtendedHeader, AppExtrinsic, AppId, BlockLengthColumns,
 	BlockLengthRows, DataProof, OpaqueExtrinsic, BLOCK_CHUNK_SIZE,
 };
-use da_runtime::{apis::DataAvailApi, Runtime, UncheckedExtrinsic};
+use da_runtime::{apis::DataAvailApi, Runtime, RuntimeCall, UncheckedExtrinsic};
 use frame_system::{limits::BlockLength, submitted_data};
 use jsonrpsee::{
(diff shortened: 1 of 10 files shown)


Leouarz (approved):

rpc/kate-rpc/src/lib.rs
```diff
@@ -454,3 +480,12 @@ fn non_extended_dimensions(ext_data: &DMatrix<BlsScalar>) -> RpcResult<Dimension
 
 	Ok(dimensions)
 }
+
+fn is_da_transaction<A, C, D: SignedExtension>(
```
fmiguelgarcia, line 484: That is NOT going to work if a Runtime Upgrade affects the DA calls. You should re-use the filters we have in Runtime, and call it through the runtime api, like `self.client.runtime_api().<filter_function>( raw_extrinsic)`

rpc/kate-rpc/src/lib.rs
```diff
@@ -418,6 +418,32 @@ where
 			.ok_or_else(|| internal_err!("Missing block hash {:?}", at))?
 			.block;
 
+		// data index starts from 0
+		let mut data_idx: i32 = -1;
```
fmiguelgarcia, line 422: I think it should be easier if you `filter` first and then use `nth` function. Something like:
```Rust
if let Some((data_idx, _ex)) = block.extrinsics().iter().enumerate()
   .filter( |(_data_idx,ex)| self.client.runtime_api().<filter_function>(&ex))
   .nth( transaction_index) {
   
      // use `data_idx` here
   }

rpc/kate-rpc/src/lib.rs
aterentic-ethernal: wdyt of using `transaction_index` in `submitted_data::calls_proof` instead of leaf index, and compute leaf index inside?

rpc/kate-rpc/src/lib.rs
0xSasaPrsic: moved logic after initial filtering when calling `calls_proof`

rpc/kate-rpc/src/lib.rs
fmiguelgarcia: I think we should move completely the Kate RPC into LC logic or/and make some utilities in `avail-subxt`. I mean there is **no needed to run this code inside a node**. att @aterentic-ethernal

aterentic-ethernal: If I'm not missing something, this has to be run by the node since LC doesn't fetch all transactions, only transactions related to specified application, which is not enough to generate proof.

aterentic-ethernal (approved):

luka-ethernal (approved):

PR #204, opened 2023-08-24
Ghali/benchmark data root

Leouarz:


ToufeeqP (approved):
LGTM

pallets/dactr/src/benchmarking.rs
```diff
@@ -204,6 +204,36 @@ benchmarks! {
 		let _data_root =submitted_data::extrinsics_root::<T::SubmittedDataExtractor, _>(once(&opaque));
 	}
 
+	// This benchmark is not directly used by extrinsic.
+	// It is mostly used to check that the weight is lower or approximately equal the `data_root` benchmark
+	data_root_batch {
+		let i in 0..(2 * 1024 * 1024);
```
fmiguelgarcia, line 210: `i` should be greater than `max_tx_size`,  and we should use the `max_block_len` (not sure about this name)
```suggestion
		let i in max_tx_size..max_block_len;
```

It would be nice if the step is `max_tx_size` too

pallets/dactr/src/benchmarking.rs
```diff
@@ -204,6 +204,36 @@ benchmarks! {
...
+
+		let max_tx_size = T::MaxAppDataLength::get();
+		let nb_full_tx = i / max_tx_size;
+		let remaining_size = i % max_tx_size;
+		let nb_additional_tx = if remaining_size > 0 { 1 } else { 0 };
+		let mut calls = Vec::with_capacity(nb_full_tx as usize + nb_additional_tx as usize);
```
fmiguelgarcia, line 216: nit:
```suggestion
		let mut calls = Vec::with_capacity(nb_full_tx as usize + 1usize);
```

pallets/dactr/src/weights.rs
```diff
@@ -145,10 +153,17 @@ impl WeightInfo for () {
...
-			.saturating_add(Weight::from_ref_time(4_321_u64).saturating_mul(i as u64))
+		// Minimum execution time: 1_457 nanoseconds.
+		Weight::from_ref_time(1_487_000_u64)
+			// Standard Error: 21
+			.saturating_add(Weight::from_ref_time(5_251_u64).saturating_mul(i as u64))
+	}
+	/// The range of component `i` is `[0, 2097152]`.
+	fn data_root_batch(i: u32, ) -> Weight {
+		// Minimum execution time: 778 nanoseconds.
+		Weight::from_ref_time(825_000_u64)
```
fmiguelgarcia, line 164: It is weird that one `data_root` takes `1_487` as minimum but several `data_root` takes `824` (less). 
I mean It should be equal or greater, right?

pallets/dactr/src/benchmarking.rs
Leouarz: Here, i go from 0 to 2mb et make n transactions from it, so it will benchmark the creation of dataroot until the maximum that we can put in a block.

Why should we go from max_tx_size to max_block_len which is basically the same from 512kb to 2mb ?

pallets/dactr/src/weights.rs
```diff
@@ -84,10 +85,17 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
 	}
 	/// The range of component `i` is `[0, 524288]`.
 	fn data_root(i: u32, ) -> Weight {
-		// Minimum execution time: 609 nanoseconds.
-		Weight::from_ref_time(672_000_u64)
-			// Standard Error: 12
-			.saturating_add(Weight::from_ref_time(4_321_u64).saturating_mul(i as u64))
+		// Minimum execution time: 1_457 nanoseconds.
+		Weight::from_ref_time(1_487_000_u64)
```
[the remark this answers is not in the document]
Leouarz: Yes i was confirmed that it's not from the changes in the algorithm.

Also benchmarking like this is quite not stable meaning we can get different result on retry.

pallets/dactr/src/weights.rs
Leouarz: I guess when the range is greater the results are smoother.

fmiguelgarcia (approved):

PR #206, opened 2023-08-25
Disable DA checks on fast syncs

fmiguelgarcia:
**DO NOT merge**: It needs some research on validator side-effects.
Disable DA Protocol checks on fast sync, like `--sync warp/fast/fast-unsafe`

PR #208, opened 2023-08-28
add --unsafe-da-sync flag to quickly sync the chain

Leouarz:
This needs to wait for the sync to get fixed on develop.

closes availproject/engineering#238


Leouarz:
> I think this clash with this one #206. I would rather use the **standards flags** `--sync full/fast/fast-unsafe/warp` (like PR#206) and add the `block.origin` check you got here. I mean `warp` neither `fast` would work as expected if DA checks are not disabled.

I think warp sync can work with da_check. I mean the flag gives more flexibility than disabling it with every sync. Also it's not compatible with validators and prompts a correct error.

Wdy think @fmiguelgarcia @prabal-banerjee

markopoloparadox (approved):
LGTM

PR #211, opened 2023-08-28
feat: fix bugs, optimize gas, add attribution

QEDK:
* fix invalid index bug
* remove redundant variables
* optimize some gas
* convert contract to interface
* remove extraneous function
* add attribution
* update to use open-source license

### DIFF
--- a/examples/validium/contracts/ValidiumContract.sol
+++ b/examples/validium/contracts/ValidiumContract.sol
@@ -1,46 +1,38 @@
-//SPDX-License-Identifier: UNLICENSED
-pragma solidity 0.8.19;
+// SPDX-License-Identifier: Apache-2.0
+// Modified from https://github.com/QEDK/solidity-misc/blob/master/contracts/Merkle.sol
+pragma solidity ^0.8.21;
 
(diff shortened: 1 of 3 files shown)


examples/validium/contracts/ValidiumContract.sol
```diff
@@ -58,10 +50,9 @@ contract ValidiumContract is Ownable {
                     }
                 }
             }
-            isMember := eq(leaf, rootHash)
-
+            // checks if the calculated root matches the expected root
+            // then, check if index was zeroed while calculating proof, else an invalid index was provided
+            isMember := and(eq(leaf, rootHash), iszero(index))
```
0xSasaPrsic, line 55: Does this work with odd number of leaves? will index be 0?

QEDK: It will work any number of leaves yes, we just validate that index is 0 at the end, otherwise it means an invalid index (i.e. index > width) was passed.

markopoloparadox (approved):
LGTM

0xSasaPrsic (approved):

PR #212, opened 2023-08-31
Add Tx Index on submit data

fmiguelgarcia:
### Description
It **adds Tx index** into the Block DB and the **event only** notifies the **hash of the data**

### New fees
- Adding the Tx index adds **1 DB Read extra cost**.
- The hash of the data **increase the base cost** of the extrinsic **and the cost per byte**.


pallets/dactr/src/tests.rs
```diff
@@ -83,10 +84,14 @@ mod submit_data {
...
 			let data = AppDataFor::<Test>::try_from(vec![b'X'; max_app_key_length]).unwrap();
+			let data_hash = H256(sp_io::hashing::blake2_256(&data));
 
-			assert_ok!(DataAvailability::submit_data(alice, data.clone()));
+			assert_ok!(DataAvailability::submit_data(alice, data));
 
-			let event = RuntimeEvent::DataAvailability(Event::DataSubmitted { who: ALICE, data });
+			let event = RuntimeEvent::DataAvailability(Event::DataSubmitted {
+				who: ALICE,
+				data_hash,
```
aterentic-ethernal, line 93: Is this a breaking change, are events part of a public interface?

aterentic-ethernal (approved):
Code looks good from my side, I'm not sure about feature that is added (we can now get index by extrinsic hash and use this in RPC we want to have?). We should just check if this is not breaking change. For trade offs,  i will rely on other approvers for better context :)

pallets/dactr/src/tests.rs
Leouarz: I don't think events are used anywhere

Leouarz (approved):

pallets/dactr/src/lib.rs
```diff
@@ -130,7 +132,21 @@ pub mod pallet {
...
+
+			// SAFETY: `data.len()` is always less than `u32::MAX` because it is bounded by
+			// `BoundedVec`
+			let len = data.len() as u32;
+
+			// Index Tx in DB block.
+			let data_hash = blake2_256(&data);
+			let extrinsic_index = <frame_system::Pallet<T>>::extrinsic_index()
+				.ok_or_else(|| Error::<T>::BadContext)?;
+			transaction_index::index(extrinsic_index, len, data_hash);
```
prabal-banerjee, line 144: For my understanding, where is this index used? Is this unique? Can I use this to query my txn?

kroos47 (approved):

pallets/dactr/src/lib.rs
```diff
@@ -130,7 +132,21 @@ pub mod pallet {
 		) -> DispatchResultWithPostInfo {
 			let who = ensure_signed(origin)?;
 			ensure!(!data.is_empty(), Error::<T>::DataCannotBeEmpty);
-			Self::deposit_event(Event::DataSubmitted { who, data });
+
+			// SAFETY: `data.len()` is always less than `u32::MAX` because it is bounded by
+			// `BoundedVec`
+			let len = data.len() as u32;
```
i1i1, line 138: `as u32` never panics and therefore safe. You should try:
```rust
let len = u32::try_from(data.len()).expect("`data` has bounded length therefore < u32::MAX");
```

pallets/dactr/src/lib.rs
fmiguelgarcia: - The index is used internally in the DB. The bug is that we was merging Indexed (regular Tx) & Non Indexed (our submit_data tx) in the same block.
- Yes, it is unique. Same code is used in other non-state-change extrinsics like `Remark/store` transaction in `substrate/frame/remak/lib.rs`.
- It is used internally, but I did not see any public API/SDK yet (at least in v0.9.37). Most likelihood we'll see it in the feature.
att. @prabal-banerjee

pallets/dactr/src/lib.rs
fmiguelgarcia: Hi @i1i1 
As you know in some platforms `usize >> u32`, and that is why we usually use `u32::try_from`. However, as: 
- `data` is a `BoundedVec<T,S>` where `S: Get<u32>` and
- `Config::MaxAppDataLenght` is also `Get<u32>`, 
we are sure that `data.len` can be safely converted to `u32` without extra checks at runtime.

Leouarz (commented):
@fmiguelgarcia Why should we switch to hash in the event if it does not improve performances ?

PR #215, opened 2023-09-04
Marko/update governance model

markopoloparadox:
## Short Description
This PR adds a new pallet (Mandate) that allows the Technical Committee to execute privileged operations.

## Long Description
The goal of this task was to implement a solution that would enable us to execute privileged operations through multiple simple-to-understand layers. So far, there were no complaints about how we use the sudo pallet, but understanding and reasoning with the democracy pallet proved to be a cognitive burden for most of us. Therefore, the decision was made to "replace" it with something simpler and easier to use.

In this case, the approach was to reuse the existing Technical Committee collective and pair it with a new custom pallet that extends the TC's abilities to execute privileged calls.

## Usage
When creating a TC proposal, make sure that the outermost extrinsic is the mandate one from the Mandate pallet. This will ensure that the caller's origin will be changed to Root when the proposal is executed.

## Technical Reasoning behind this change
Here is a list of other rejected potential implementations:

- Custom Sudo Pallet: Forking the Sudo pallet just to add one small feature adds technical debt to our system.
- Custom Technical Committee Pallet: Same as with the previous idea, forking just to make a small tweak isn't worth it.
- Moving to Democracy v2: This would make our system even more complex and harder to reason with. Since we have no user base, there would be no benefits in having this kind of governance.


Leouarz (approved):
Seems familiar, SUS

pallets/mandate/src/lib.rs
```diff
@@ -0,0 +1,75 @@
...
+		/// A sudo-able call.
+		type RuntimeCall: Parameter
+			+ UnfilteredDispatchable<RuntimeOrigin = Self::RuntimeOrigin>
+			+ GetDispatchInfo;
+
+		/// Type representing the weight of this pallet
+		type WeightInfo: WeightInfo;
+
+		/// Someone who can call the mandate extrinsic.
+		type ExternalOrigin: EnsureOrigin<Self::RuntimeOrigin>;
```
ToufeeqP, line 38: `ApprovedOrigin` would be a better name I guess, but fine with this as well

ToufeeqP (approved):

pallets/mandate/src/lib.rs
markopoloparadox: Good idea, thank you!

PR #216, opened 2023-09-06
Updated RPC Response Field Names to Use Camel Case

ToufeeqP:
Note: This change is not compatible with the Light Client

avail-core companion: [PR#45](https://github.com/availproject/avail-core/pull/45)
Part of: [Issue #250](https://github.com/availproject/engineering/issues/250)

### DIFF
--- a/avail-subxt/build_api.sh
+++ b/avail-subxt/build_api.sh
@@ -35,6 +35,8 @@ subxt codegen \
 	--derive-for-type avail_core::AppId=derive_more::From \
 	--url http://localhost:9933 \
 	| sed -En "s/pub struct KateCommitment/#\[serde\(rename_all = \"camelCase\"\)\] \0/p" \
+	| sed -En "s/pub struct HeaderExtension/#\[serde\(rename_all = \"camelCase\"\)\] \0/p" \
+	| sed -En "s/pub struct DataLookupItem/#\[serde\(rename_all = \"camelCase\"\)\] \0/p" \
 	| sed -E '1i \#\[allow(clippy::all)]' \
 	| rustfmt --edition=2021 --emit=stdout > src/api_dev.rs
 echo "🎁 Avail-SubXt API generated in 'src/api_dev.rs'"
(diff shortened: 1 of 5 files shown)


avail-subxt/build_api.sh
```diff
@@ -35,6 +35,8 @@ subxt codegen \
 	--derive-for-type avail_core::AppId=derive_more::From \
 	--url http://localhost:9933 \
 	| sed -En "s/pub struct KateCommitment/#\[serde\(rename_all = \"camelCase\"\)\] \0/p" \
+	| sed -En "s/pub struct HeaderExtension/#\[serde\(rename_all = \"camelCase\"\)\] \0/p" \
+	| sed -En "s/pub struct DataLookupItem/#\[serde\(rename_all = \"camelCase\"\)\] \0/p" \
```
i1i1, line 39: I think you might wanna use `subxt generate --attributes-for-type DataLookupItem="#[serde(rename_all = \"camelCase\")]"` for all of those. Found it in sources:
https://github.com/paritytech/subxt/blob/master/cli/src/commands/codegen.rs#L34-L35

Or might be a good idea to enforce this rule on all the types and just add `--attributes "#[serde(rename_all = \"camelCase\")]"` for that matter :man_shrugging:

ToufeeqP: Thank you for the suggestion. I think we've already explored the `--attributes-for-type` option, but it didn't work for our specific use case. And we don't intend to apply the `#[serde(rename_all = \"camelCase\")]` attribute to all types, but only to those that appear in the RPC responses. Therefore, we've implemented this workaround using `sed` to achieve the desired result

i1i1: Out of curiosity, what was the reason it didn't do the job?

ToufeeqP: No idea

markopoloparadox (approved):

avail-subxt/build_api.sh
i1i1: Looked into it. This version of subxt doesn't have this flag. If we update subxt library and binary then those sed can be removed, I think.

Could you add TODO for that?

ToufeeqP: It's there in the v0.29.0:  https://github.com/paritytech/subxt/blob/e40a8629e279e80a7fbb56ff553a430a36612956/cli/src/commands/codegen.rs#L34

PR #217, opened 2023-09-07
Add semver version constants to DA Runtime for LC compat tracking

luka-ethernal:
Comments welcome if you have better ideas for naming, automation, etc.

### DIFF
--- a/pallets/dactr/src/lib.rs
+++ b/pallets/dactr/src/lib.rs
@@ -61,6 +61,18 @@ pub mod pallet {
 		#[pallet::constant]
 		type MaxAppDataLength: Get<u32>;
 
+		/// Major version of runtime
+		#[pallet::constant]
+		type AvailRuntimeVersionMajor: Get<u16>;
+
+		/// Minor version of runtime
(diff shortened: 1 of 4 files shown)


i1i1:
Can't LC just use `RuntimeVersion` for that?

https://github.com/availproject/avail/blob/a19d6e6ed9e27973c13d7fa82367996bacce2898/runtime/src/lib.rs#L120-L133

luka-ethernal:
> Can't LC just use `RuntimeVersion` for that?
> 
> https://github.com/availproject/avail/blob/a19d6e6ed9e27973c13d7fa82367996bacce2898/runtime/src/lib.rs#L120-L133

As discussed a few weeks ago, we decided that runtime version parameters (ie spec_version_impl_version) have special meaning, which is meant for nodes interpretation, and are not suitable to be interpreted by LC to decide the compatibility and we concluded we are going to add a semver set of our own specifically to track the version and let LC use it to determine the compatibility.

i1i1 (approved):

PR #218, opened 2023-09-08
Upgrade/polkadot v1.0

Leouarz:


runtime/src/migration.rs
```diff
@@ -138,5 +251,53 @@ mod nomination_pools {
...
+				storage_version
+			);
+			Ok(sp_std::vec![])
+		}
+
+		// Here we update the storage version to be consistent even if pallet was initialized correctly
+		pub fn on_runtime_upgrade() -> Weight {
+			let storage_version = NominationPools::on_chain_storage_version();
+			if storage_version < 3 {
+				StorageVersion::new(3).put::<NominationPools>();
```
ToufeeqP, line 287: Essentially, for the Kate testnet, do we only need to increase the version to 3 without performing any actual data migration?

runtime/src/migration.rs
```diff
@@ -25,110 +25,223 @@ use sp_runtime::TryRuntimeError;
...
+		}
 	}
+}
 
-	mod v0_to_v1 {
+mod nomination_pools {
+	use super::*;
+
+	#[allow(dead_code)]
+	pub mod v0_to_v1 {
```
ToufeeqP, line 205: Do we need this? I think once both our networks are in sync, we can just clear the upstream pallet migrations from our code.

ToufeeqP (approved):

runtime/src/migration.rs
Leouarz: yes, the data is already in the correct format, it's just a storage version mismatch after all. THe v3 to v5 will be done after storage version update

runtime/src/migration.rs
Leouarz: It's good practice to keep it as parity is doing it i guess, but it's not used indeed

ToufeeqP: What I'm trying to convey is that for pallets that are part of the upstream, we can remove them once there are no more migrations required on any chain. However, for our own pallets, we can choose to keep them.

Leouarz: Agreed, i was just wondering if there was any reason to keep it ? (did not really touch it since it's not my code)
Any thought @MiguelDD1

PR #219, opened 2023-09-12
update to weightv2, remove warning, use weights folder in runtime lik…

Leouarz:
…e polkadot-sdk

**The base branch will be changed to develop when toufeeq/polkadot-v1.0 will be merged**


.maintain/frame-weight-template.hbs
```diff
@@ -29,7 +31,7 @@ pub trait WeightInfo {
 	{{/each}}
 }
 
-/// Weights for {{pallet}} using the Data Avaiability node and recommended hardware.
+/// Weights for `{{pallet}}` using the Substrate node and recommended hardware.
```
ToufeeqP, line 34: Should we use Data Availability instead of Substrate in the template?

Leouarz: Yeah this is just a copy paste😅

README.md
```diff
@@ -82,13 +82,27 @@ running the benchmarks from `da-control` pallet, and the generated file is
         --steps=30 \
         --repeat=20 \
         --log=warn \
-        --execution=wasm \
-        --wasm-execution=compiled \
```
ToufeeqP, line 86: Shouldn't we use these execution strategies for benchmarking?

markopoloparadox: They are not needed anymore :pepe_salute:

Leouarz: Yep they are "used" by default now.

--execution was removed (doc is not yet updated)
--wasm-execution compiled is default

runtime/src/weights/pallet_dactr.rs
```diff
@@ -0,0 +1,115 @@
...
+
+// Executed Command:
+// ./target/release/data-avail
+// benchmark
+// pallet
+// --chain=dev
+// --steps=50
+// --repeat=25
+// --log=warn
+// --execution=native
```
ToufeeqP, line 33: I think we should generate the weights on wasm execution(worst case)

Leouarz: I did not rerun benchmarks yet, it will be done on a separate PR sor this is the old one, you can see we don't have proof size on this file.

Leouarz: What do you mean by ? I think we should generate the weights on wasm execution(worst case)

ToufeeqP (approved):

ToufeeqP:
> Love it, 5/7

If you love, why 5 then😜, Am I missing some context here

Leouarz:
> > Love it, 5/7
> 
> If you love, why 5 then😜, Am I missing some context here

https://www.dictionary.com/e/memes/5-7/

PR #242, opened 2023-09-19
Marko/adding more scripts

markopoloparadox:
---
Name: Pull Request
About: Provide information about the proposed changes
---
# Pull Request type

- [ ] Feature
- [ ] Bugfix
- [ ] Refactor
- [ ] Format
- [ ] Documentation
- [ ] Testing
- [X] Other: New Scripts

## Description
Added script to fork our live networks and added script to compare our live networks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,8 +186,44 @@ This describes the different sync modes available for the substrate chain and th
 | warp               | partially compatible       | compatible                                                                         |
 | fast / fast unsafe | not compatible             | compatible (with [warnings](https://github.com/paritytech/polkadot-sdk/issues/19)) |
 
+## Forking Kate/Biryani/Dymension
 
+### Overview
+The following script can be used to fork our live networks:
 
+```bash
+# kate
+CONFIG=./scripts/networks/kate.yaml ./get_state.sh
+
+# Biryani
+CONFIG=./scripts/networks/biryani.yaml ./get_state.sh
+
+# Dymension
+CONFIG=./scripts/networks/dymension.yaml ./get_state.sh
+```
+
+Once the script is done, the forked chain (together with raw data) can be found
+inside the output folder. To use the forked chain run the following command:
+
+```bash
+cargo run -- --chain ./output/fork.json --alice --tmp
+```
+
+### Custom Binary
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -186,8 +186,44 @@ This describes the different sync modes available for the substrate chain and th
 | warp               | partially compatible       | compatible                                                                         |
 | fast / fast unsafe | not compatible             | compatible (with [warnings](https://github.com/paritytech/polkadot-sdk/issues/19)) |
 
+## Forking Kate/Biryani/Dymension
```
ToufeeqP, line 189: Do we want to expose the Biryani/Dymension networks as well? @markopoloparadox @Leouarz

Leouarz: Not sure, but there won't be validators nor funds for user to use the network, so it's ok I guess.

README.md
```diff
@@ -186,8 +186,44 @@ This describes the different sync modes available for the substrate chain and th
...
+version. The script accepts custom binary paths.
+
+```bash
+CONFIG=./scripts/networks/kate.yaml BINARY=./target/release/data-avail ./get_state.sh
+```
+
+## Compare two running chains
+You can compare two chains by executing the following scripts:
+```bash
+deno run --allow-net --allow-write  ./scripts/compare.ts ws://127.0.0.1:9944  wss://kate.avail.tools/ws
```
ToufeeqP, line 223: Is this the same script you used to compare the changes in upgrade v1.0 PR? @markopoloparadox

README.md
markopoloparadox: @ToufeeqP Will ask Jakub what he thinks

README.md
markopoloparadox: Correct

ToufeeqP: btw, it was not accurate🧐

markopoloparadox: I remember there was a chance that some information was missing from the report. I will monitor this and adjust the script if we face a similar problem next time. Besides that (missing data), the report cannot be incorrect since it compares two metadatas and does a diff, ie it cannot hallucinate changes that are not there.

PR #247, opened 2023-09-22
Format and Sort features in `Cargo.toml` files

ToufeeqP:
# Pull Request type

- [ ] Feature
- [ ] Bugfix
- [ ] Refactor
- [x] Format
- [ ] Documentation
- [ ] Testing
- [x] Other: CI update

## Checklist
- [x] The tests pass succesfully with `cargo test`.


fmiguelgarcia (commented):
Could you pls add the format checker to the CI?
Are you using `dprint` for this?

ToufeeqP:
> Could you pls add the format checker to the CI? Are you using `dprint` for this?

I am using Zepter, which parity also uses. Was thinking of adding it to the formatting but wasn't sure

markopoloparadox (approved):

fmiguelgarcia:
> s thinking of adding it to the formatting bu

fmiguelgarcia:
> > Could you pls add the format checker to the CI? Are you using `dprint` for this?
> 
> I am using [Zepter](https://lib.rs/crates/zepter), which parity also uses. Was thinking of adding it to the formatting but wasn't sure, I will add it now

In `avail-core` we are using `dprint` which is also integrated in its CI, and format also `JSON` & `MarkDown` files.
We should unify both repos.

ToufeeqP:
> > > Could you pls add the format checker to the CI? Are you using `dprint` for this?
> > 
> > 
> > I am using [Zepter](https://lib.rs/crates/zepter), which parity also uses. Was thinking of adding it to the formatting but wasn't sure, I will add it now
> 
> In `avail-core` we are using `dprint` which is also integrated in its CI, and format also `JSON` & `MarkDown` files. We should unify both repos.

Ok, I didn't noticed that we already have a toml formatting tool integrated. And, I agree that we should use the single one in both repos. so shall we use `dprint` only then?

.github/workflows/default.yml
```diff
@@ -45,6 +45,22 @@ jobs:
...
           args: --manifest-path avail-subxt/Cargo.toml --check
 
+      - name: Install zepter
+        continue-on-error: true
+        run: |
+          cargo install zepter --locked --version 0.11.0 -q -f --no-default-features && zepter --version
+
+      - name: Check Avail toml formatting
+        # Not enforcing the toml formatting for now
+        continue-on-error: true
```
fmiguelgarcia, line 55: Not enforcing?

ToufeeqP: Yeah, the check is just for awareness purposes now, later maybe we can enforce it, wdyt? just to not annoy the devs on toml formatting

fmiguelgarcia: IMHO we should enforce it. Otherwise we will need a periodically task to apply formatting.

fmiguelgarcia (approved):

ToufeeqP:
> @ToufeeqP What do you think about moving lint related checks into a new workflow? Should make it easier to reason with what our workflows are actually doing.

It makes sense to me

PR #251, opened 2023-10-03
Header Generation Benchmark in CI

fmiguelgarcia:
# Pull Request type

- [ ] Feature
- [ ] Bugfix
- [ ] Refactor
- [ ] Format
- [ ] Documentation
- [ ] Testing
- [x] Other: CI

## Description

It adds a couple of performance benchmarks to the CI, so **any performance regression greater than 15%** will be marked as a failure. The initial benchmarks are measuring the _Kate commitment generation_ over different number of columns (like, `32`, `64`,...`256`).
There are 2 methods for performance measurements:
- Using [Criterion](https://bheisler.github.io/criterion.rs/book/criterion_rs.html) benchmarks, running on our AWS reference machine.
- Using [IAI](https://bheisler.github.io/criterion.rs/book/iai/iai.html), which is an _" experimental benchmarking harness that uses Cachegrind to perform extremely precise single-shot measurements ... it's useful for reliable benchmarking in CI"_

## Checklist
- [x] The tests pass succesfully with `cargo test`.


markopoloparadox:
@fmiguelgarcia 
Not sure if we should use iai.
The project is not maintained anymore and it doesn't work with newer versions on valgrind (>3.18)
https://github.com/bheisler/iai/issues/37
https://github.com/bheisler/iai/issues/36

Is there anything wrong with Criterion?

fmiguelgarcia:
As you suggested, I've also added `Criterion` bench-marking. It is executed inside our AWS _reference_ machine, in order to limit divergences between executions.
I initially added IAI because its execution using `Cachegrind` is perfect for the CI due to the stability of the measurements, independently of the current load. Please take a look into [Criterion doc about IAI](https://bheisler.github.io/criterion.rs/book/iai/iai.html).
In any case, the benchmark github action supports several benchmarking tools, and it really easy to add new ones. 
As example, we could add substrate weight benchmarking too, so it would warranty no performance regression in the cost of extrinsics between releases.

att. @markopoloparadox 
> @fmiguelgarcia Not sure if we should use iai. The project is not maintained anymore (last commit is from 2021) and it doesn't work with newer versions on valgrind (>3.18) [bheisler/iai#37](https://github.com/bheisler/iai/issues/37) [bheisler/iai#36](https://github.com/bheisler/iai/issues/36)
> 
> Is there anything wrong with Criterion?
> 
> Besides this, I got the feeling that we should have two separate tasks. One for implementing the benchmarks and the second one for the CI (which would include running the benchmarks and store the results inside a db)

markopoloparadox (approved):

PR #253, opened 2023-10-05
Added Deno examples and benchmarks

markopoloparadox:


examples/deno/benchmarks/query_proof.ts
```diff
@@ -0,0 +1,63 @@
...
+const promises = [];
+console.log("Querying 8.5k cells...")
+performance.mark("start");
+for (let counter = 0; counter < count; counter += 30) {
+    end = counter + 30;
+    end = end > count ? count : end;
+    promises.push(api.rpc.kate.queryProof(cells.slice(counter, end), targetBlockHash))
+}
+
+await Promise.all(promises);
```
ToufeeqP, line 59: So we're considering the happy path here? what if some requests return a reject, won't it give inaccurate results? Maybe we can use `Promise.allSettled` instead of `Promise.all`?

ToufeeqP (approved):

fmiguelgarcia (approved):

Leouarz (approved):

PR #299, opened 2023-11-10
Removed Kate, Dymension, and Biryani from cmd list

markopoloparadox:


node/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "data-avail"
-version = "1.8.0"
+version = "1.8.1"
```
ToufeeqP, line 3: Should we update the versioning scheme here as well? cc: @markopoloparadox @Leouarz

ToufeeqP (approved):

node/Cargo.toml
markopoloparadox: Possibly. Let's discuss this the next time we see each other :)

PR #322, opened 2023-12-04
modify the dockerfile to use correct versions and tags, and install curl

soragXYZ:
# Pull Request type

- [ ] Feature
- [ ] Bugfix
- [ ] Refactor
- [ ] Format
- [ ] Documentation
- [ ] Testing
- [X] Other: labels + tools

## Description

I have updated the label of the dockerfile since they were outdated. But I don't know if it was the correct place to do it.  
Also, I have added curl to the image: I've just set up my validator according to your documentation, and in [this part](https://docs.availproject.org/operate/validator/staking/#rotating-session-keys), we need to have curl installed before running
```bash
docker exec -i da097bbeba75 curl -sH "Content-Type: application/json" -d '{"id":1, "jsonrpc":"2.0", "method": "author_rotateKeys", "params":[]}' http://localhost:9944
```
So it can be usefull for future validator to have curl directly embedded in the image, or if you refuse this PR, can you update your gitbook to reflect this change with this command ?
```bash
docker exec -i <CONTAINER_ID> apt install curl -y
```

## Checklist
- [X] I have performed a self-review of my own code.
- [ ] The tests pass successfully with `cargo test`.
- [ ] The code was formatted with `cargo fmt`.
- [X] If this change affects documented features or needs new documentation, I have created a PR with a [documentation update](https://github.com/availproject/availproject.github.io).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@ RUN apt update -y && \
 WORKDIR "/da/src"
 
 # Clone repo
-ARG AVAIL_TAG=v1.8.0.2
+ARG AVAIL_TAG=v1.8.0.3
 RUN git clone -b $AVAIL_TAG --single-branch https://github.com/availproject/avail.git .
 
 # This installs Rust and updates Rust to the right version.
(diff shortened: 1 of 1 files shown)


soragXYZ:
Here is the PR in the documentation guide for reference: [here](https://github.com/availproject/availproject.github.io/pull/175)

Feel free to close this one, or the on the other repo, since they are mutually exclusive

PR #330, opened 2023-12-06
fix typo in avail-subxt/src/api_dev.rs

Ursulafe:
minor typo fix

### DIFF
--- a/avail-subxt/src/api_dev.rs
+++ b/avail-subxt/src/api_dev.rs
@@ -5251,10 +5251,10 @@ pub mod api {
 			use super::runtime_types;
 			pub struct ConstantsApi;
 			impl ConstantsApi {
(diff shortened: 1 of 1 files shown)


avail-subxt/src/api_dev.rs
```diff
@@ -5251,10 +5251,10 @@ pub mod api {
 			use super::runtime_types;
 			pub struct ConstantsApi;
 			impl ConstantsApi {
-				#[doc = " A fee mulitplier for `Operational` extrinsics to compute \"virtual tip\" to boost their"]
```
markopoloparadox, line 5254: These comments are autogenerated and taken directly from substrate. Manually changing them won't do anything since it will be replaced the next time we call the build_api.sh script.

PR #342, opened 2023-12-13
Correcting all the grammatical errors!

vehorny:
Hello I am learning web3 programming, I went through some of the code because I was told modulatiry is fascinating (and it is). I found some little errors in comments, nothing important, except one of them that is printed to the user. I checked all files so the repo should now be typo free! I hope you don't mind opening a PR for such a little thing, I read the contributing file and didn't find it being forbidden, but even if it is little I hope i'll help you even just a bit in your research. Thank you for making that repo public. Have a nice day.

### DIFF
--- a/examples/deno/query_proof.ts
+++ b/examples/deno/query_proof.ts
@@ -18,7 +18,7 @@ const submitData = new Promise<[H256, H256]>((res, _) => {
     });
 });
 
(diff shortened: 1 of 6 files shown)


markopoloparadox:
This look like a PR made from a bot account.

PR #343, opened 2023-12-14
Correcting all typo errors.

vehorny:
Hi, 

This is a following of #342 
I am not a bot. This is an implementation of the python codespell library. 
I get that having many typos might be hurtfull as it can lead to spam, that is why I corrected all the typos in one go.

This will prevent you from being spammed over and over with typos PR in the future since there won't be any left.

Have a nice evening


markopoloparadox:
This look like a PR made from a bot account.

PR #353, opened 2023-12-27
Added special segment to readme about how we treat external pull requests

markopoloparadox:
We are getting targeted with bots trying to "fix" grammatical errors. This is most likely the after-effect of Celestia airdrop where you would get more tokens if you had more commits it blockchain like projects. 

The bots that we are dealing with are obnoxious and keep creating new PRs until we finally ban them. 
Examples:
  
Example 1: User Vehorny
https://github.com/availproject/avail/pull/342
https://github.com/availproject/avail/pull/343
https://github.com/availproject/avail/pull/344

Example 2: User JohnwickBC
https://github.com/availproject/avail/pull/337
https://github.com/availproject/avail/pull/338

Example 3: User th3nyst1c
https://github.com/availproject/avail/pull/349
https://github.com/availproject/avail/pull/351

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,27 +1,3 @@
 # Contributing to Avail
 
-Thank you for your interest in contributing to Avail! Your contributions are greatly appreciated, and this guide will provide you with all the information you need to get started.
-
-## Code of Conduct
-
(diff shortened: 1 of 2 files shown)


th3nyst1c:
Now I understand why you thought I was a bot. You can tell right away. However, if I were you, I would include in the airdrop only those users who contributed a high number of commits and changes. There's no point in calling users "bots".

PR #368, opened 2024-01-12
add remark-with-event example to avail-js

stanlagermin:
I add an example about remark with event to avail-js.


Leouarz:
Thank you for your contribution, hower there are already extremely similar examples and extractData is not compatible with remark extrinsics but only da submission.
image

PR #403, opened 2024-02-06
Ghali/clean and update vector

Leouarz:
# Pull Request type

- [x] Refactor

## Description

- Vector bridge now use storage instead of const for step / rotate function ids and verifier key

## Checklist


0xSasaPrsic (approved):
Looks good.

pallets/vector/src/lib.rs
```diff
@@ -130,6 +133,16 @@ pub mod pallet {
 			slots_per_period: u64,
 			finality_threshold: u16,
 		},
+		/// Function Ids were updated
+		FunctionIdsUpdated { value: Option<(H256, H256)> },
```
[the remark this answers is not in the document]
Leouarz: I just replicated the nomenclature i found above, and just realized that even before those changes, in the same file, there is some "Update" and some "Updated". Anyway should be ok, it's @0xSasaPrsic's decision

pallets/vector/src/lib.rs
```diff
@@ -175,6 +188,27 @@ pub mod pallet {
...
+	/// The order of storage is (step_function_id, rotate_function_id)
+	#[pallet::storage]
+	#[pallet::getter(fn function_ids)]
+	pub type FunctionIds<T: Config> = StorageValue<_, Option<(H256, H256)>, ValueQuery>;
+
+	/// Step verification key storage
+	#[pallet::storage]
+	#[pallet::getter(fn step_verification_key)]
+	pub type StepVerificationKey<T: Config> =
+		StorageValue<_, Option<BoundedVec<u8, ConstU32<10_000>>>, ValueQuery>;
```
markopoloparadox, line 204: What are our expectations on the size of Step Verification Keys? Is 10k enough to fit current and future keys?

Leouarz: I took the same specification as stuff above. It was already a bounded vec somewhere. Right now it's roughly 3 to 4 times what's necessary. We also can update it or even use new storage and delete this one incase we don't want to do a migration for that haha.

pallets/vector/src/lib.rs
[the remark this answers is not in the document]
0xSasaPrsic: Let's make it "Updated"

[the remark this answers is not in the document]
Leouarz: So you want to change name of previous events ? Won't that be an issue with operator ?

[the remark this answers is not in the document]
0xSasaPrsic: Yes, let's make it consistent. The operator does not use events.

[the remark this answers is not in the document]
Leouarz: Ok, will you push it or shall I do it ?

pallets/vector/src/lib.rs
```diff
@@ -798,5 +888,13 @@ pub mod pallet {
 		fn is_domain_valid(domain: u32) -> bool {
 			WhitelistedDomains::<T>::get().contains(&domain)
 		}
+
+		fn get_function_ids() -> Result<(H256, H256), DispatchError> {
+			if let Some(function_ids) = FunctionIds::<T>::get() {
```
markopoloparadox, line 893: Unimportant subjective style comment: Let-else would be a better choice here.

pallets/vector/src/lib.rs
```diff
@@ -735,30 +817,38 @@ pub mod pallet {
...
 
 		/// get_verifier returns verifier based on the provided function id.
-		fn get_verifier(function_id: H256) -> Result<Verifier, Error<T>> {
-			if function_id == T::StepFunctionId::get() {
+		fn get_verifier(
+			function_id: H256,
+			step_function_id: H256,
+			rotate_function_id: H256,
+		) -> Result<Verifier, Error<T>> {
+			if function_id == step_function_id {
```
markopoloparadox, line 825: Unimportant subjective style comment: Would match work in here?

pallets/vector/src/lib.rs
```diff
@@ -644,6 +725,7 @@ pub mod pallet {
 			finalized_slot: u64,
 			cfg: Configuration,
```
markopoloparadox, line 726: Could we pass a reference instead?

pallets/vector/src/lib.rs
Leouarz: yeah i get that, i sill feel this is simple and readable tbh

pallets/vector/src/lib.rs
```diff
@@ -678,8 +761,7 @@ pub mod pallet {
 			ensure!(sc_poseidon != U256::zero(), Error::<T>::SyncCommitteeNotSet);
 
 			let input = encode_packed(sc_poseidon, attested_slot);
-			let result = Self::verified_step_call(T::StepFunctionId::get(), input, step_store)?;
-
+			let result = Self::verified_step_call(step_function_id, input, step_store)?;
```
markopoloparadox, line 764: Can `input` and `step_store` be passed as references?

pallets/vector/src/tests.rs
```diff
@@ -182,7 +179,7 @@ fn test_fulfill_step_call_proof_not_valid() {
 
 		let result = Bridge::fulfill_call(
 			RuntimeOrigin::signed(TEST_SENDER_ACCOUNT),
-			STEP_FN_ID,
+			STEP_FUNCTION_ID,
```
markopoloparadox, line 182: In general, we should use the step function id that we have stored inside our storage. There is no guarantee that `STEP_FUNCTION_ID` is equal to the value that we have stored.

In any case, no changes are needed but for future tests let's do it properly.

pallets/vector/src/weights.rs
```diff
@@ -206,6 +209,18 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
 			.saturating_add(T::DbWeight::get().reads(5_u64))
 			.saturating_add(T::DbWeight::get().writes(1_u64))
 	}
+
+	fn set_function_ids() -> Weight {
```
markopoloparadox, line 213: I will assume that we don't need to correct weights right now since we are going to regenerate everything in the next phase. It would be great if we could still test it and check if our benchmarks actually work.

markopoloparadox (approved):

PR #416, opened 2024-02-15
[ ] Format Update try-runtime.yml Himess

Himess:
# Pull Request type
[ ] Format

## Description
While "try-runtime" is generally used, where I corrected it, it was used as "Try-runtime". I think it should be like this in terms of semantic integrity.

## Testing Performed
Manuel Testing


Leouarz:
Hey, it's ok to have it uppercase as it's a title and not a command.

PR #423, opened 2024-02-19
Update README.md

kriss39:
"interract" should be "interact"  and also look for "Validitum example implementation".

I am currently running avail node and it looks smooth. i will try to test new things too


Leouarz:
This will come from a separate branch

kriss39:
> This will come from a separate branch
should i add new new?

Leouarz:
It's already there, no need

PR #425, opened 2024-02-19
add extension fallback

Leouarz:
# Pull Request type

- [x] Refactor

## Description

It adds a last resort mechanism in case, for some reason, the commitment or grid construction fails.
Companion PR in avail-core : https://github.com/availproject/avail-core/pull/74

## Related Issues

[Closes engineering#435](https://github.com/availproject/engineering/issues/435)

## Checklist

### DIFF
--- a/pallets/system/src/header_builder.rs
+++ b/pallets/system/src/header_builder.rs
@@ -2,6 +2,7 @@ use avail_core::{
 	header::HeaderExtension, traits::ExtendedHeader, AppExtrinsic, DataLookup, HeaderVersion,
 };
 use frame_support::traits::Randomness;
+#[cfg(feature = "std")]
 use kate::gridgen::EvaluationGrid;
 pub use kate::{
 	metrics::{IgnoreMetrics, Metrics},
@@ -106,13 +107,14 @@ fn corrupt_commitment(block_number: u32, commitment: &mut Vec<u8>) {
 	}
 }
 
+#[cfg(feature = "std")]
 pub fn build_grid(
 	app_extrinsics: &[AppExtrinsic],
 	block_length: BlockLength,
 	seed: Seed,
 ) -> Result<EvaluationGrid, String> {
 	const MIN_WIDTH: usize = 4;
-	let grid = kate::gridgen::EvaluationGrid::from_extrinsics(
+	let grid = EvaluationGrid::from_extrinsics(
 		app_extrinsics.to_vec(),
 		MIN_WIDTH,
 		block_length.cols.0.saturated_into(), // even if we run on a u16 target this is fine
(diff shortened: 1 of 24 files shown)


pallets/system/src/header_builder.rs
```diff
@@ -104,51 +108,129 @@ fn corrupt_commitment(block_number: u32, commitment: &mut Vec<u8>) {
...
+
+	// Build the grid
+	let timer = std::time::Instant::now();
+	let maybe_grid = build_grid(app_extrinsics, block_length, seed);
+	// Evaluation Grid Build Time Metrics
+	HeaderExtensionBuilderMetrics::observe_evaluation_grid_build_time(timer.elapsed());
+	let grid = match maybe_grid {
+		Ok(res) => res,
+		Err(message) => {
+			log::error!("NODE_CRITICAL_ERROR_001 - A critical error has occured: {message:?}.");
```
prabal-banerjee, line 209: Can we have some other error code? Looks verbose. Something like this (although not equivalent)?
https://github.com/availproject/avail/wiki/Transaction-Custom-IDs

prabal-banerjee: And then a corresponding error on the wiki,

prabal-banerjee (approved):

pallets/system/src/header_builder.rs
Leouarz: We chose to not introduce the same type of error for multiple reasons : 
- This is a console only error
- This is not a transaction error that need to be returned to the user so we cannot use the same feat
- It needs to be easily findable

markopoloparadox (approved):

markopoloparadox (approved):

ToufeeqP (approved):

sh3ll3x3c (approved):

Aphoh (changes requested):
Looks pretty good! Just one more place it would be good to do an error check instead of a panic. It would be really nice to put something like [this](https://github.com/dtolnay/no-panic) in there too, but my guess is that my code downsteam doesn't pass this kind of scrutiny 😅

pallets/system/src/header_builder.rs
```diff
@@ -104,51 +108,129 @@ fn corrupt_commitment(block_number: u32, commitment: &mut Vec<u8>) {
...
+		.map_err(|e| format!("Make polynomial grid failed: {e:?}"))?;
+
+	let extended_grid = poly_grid
 		.extended_commitments(&*PMP, 2)
-		.expect("Extended commitments cannot fail")
+		.map_err(|e| format!("Grid extension failed: {e:?}"))?;
+
+	let commitment = extended_grid
 		.iter()
 		.flat_map(|c| c.to_bytes().expect("Commitment serialization cannot fail"))
```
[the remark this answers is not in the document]
Leouarz: It can be done, but this literally should never fail, wdy think ?

PR #450, opened 2024-03-04
remove custom error in favor  of exhaust error

Leouarz:
# Pull Request type

- [x] Bugfix

## Description
Currently, when the padded block length is exceeded, we actually throw away the transaction instead of returning it back to the pool. That's due to the fact that we return a custom error and not an exhaust error see the code here : 

- https://github.com/paritytech/polkadot-sdk/blob/master/substrate/client/basic-authorship/src/basic_authorship.rs#L519

It mean that everytime until now, when we see padded block length exceeded, transactions were dropped entirely !

Companion PR in avail-core : https://github.com/availproject/avail-core/pull/78

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -917,7 +917,7 @@ dependencies = [
 [[package]]
 name = "avail-core"
 version = "0.5.1"
-source = "git+https://github.com/availproject/avail-core?branch=main#861f35a79a6c278bfe68b34d1a1d2794f7734f47"
+source = "git+https://github.com/availproject/avail-core?branch=ghali/remove-old-error#2ac4494602b0c31783f405a63001565e9af89495"
 dependencies = [
  "binary-merkle-tree",
  "derive_more",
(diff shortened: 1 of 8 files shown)


markopoloparadox (approved):

ToufeeqP (approved):

prabal-banerjee (approved):
Nice catch!
- Good to update the [wiki](https://github.com/availproject/avail/wiki/Transaction-Custom-IDs)
- Would be nice to throw some debug log about existing error before throwing the exhausting of resources error

Leouarz:
> Nice catch!
> 
> * Good to update the [wiki](https://github.com/availproject/avail/wiki/Transaction-Custom-IDs)
> * Would be nice to throw some debug log about existing error before throwing the exhausting of resources error

> Done
> The log is always there

fmiguelgarcia (approved):
Great catch!

PR #493, opened 2024-03-22
Ghali/test matrix weights

Leouarz:
# Pull Request type

- [x] Feature
- [x] Refactor

## Description

This PR aims to introduce a better weights system now that only da submission go in the matrix. It use the scale encoded len to compute the number of scalars that wil goes in the matrix and make a ratio with the max scalars. We multiply this ratio with the maximum normal dispatch weight in a block.

With this, by sending roughly 2mb of data, you will fill out the whole matrix.

Now this will increase cost a lot for da_submission. It should be reduced by answering the question, how much should it cost to fill a block.

The cost for other tx will be low but it's ok since it will respect real computation.

Most important part is here : https://github.com/availproject/avail/pull/493/files#diff-deee4784b1fa5b7b7293815ce07991b7e2010c07d492094b728eb0ebb0820457R369

Please note that we have now the regular NORMAL_DISPATCH_RATIO used for block_length and block_weight. But we also have the DA_DISPATCH_RATIO, which is 100%, used by the matrix. Since only normal extrinsic / da submissions only can go in, we can fill 100% of the matrix. Filling 100% of the matrix will fill 85% of the block weight.

## Testing Performed

I managed to fill the weights correctly to 84.xx% of the 85% available. more would trigger the check_weight and we have the check_app_id (padded_len) as complement.

## Checklist

resolves https://github.com/availproject/engineering/issues/546


Leouarz:
Please wait for all reviewers to approve before merging

markopoloparadox (approved):

ToufeeqP (approved):

PR #495, opened 2024-03-23
Added extension check that returns an error if submit data or send me…

markopoloparadox:
Companion PR: https://github.com/availproject/avail-core/pull/83
resolves https://github.com/availproject/engineering/issues/540

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -875,7 +875,7 @@ dependencies = [
 [[package]]
 name = "avail-core"
 version = "0.6.0"
-source = "git+https://github.com/availproject/avail-core?branch=main#424ee429135dae04be5150db0699be4e9cd4892b"
+source = "git+https://github.com/availproject/avail-core?branch=main#faafcab4828ca0fe2d37ed62e9d1b1ff74d830d6"
 dependencies = [
  "binary-merkle-tree",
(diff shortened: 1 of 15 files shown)


Leouarz:
Isn't this possible without adding a new extension ?

pallets/transaction-payment/src/lib.rs
```diff
@@ -1048,3 +1048,13 @@ where
 		Self::compute_actual_fee(len, &info, &post_info, Zero::zero())
 	}
 }
+
+pub struct TestLengthToFeeU64;
```
Leouarz, line 1052: Why is this needed ?

ToufeeqP (approved):

pallets/transaction-payment/src/lib.rs
markopoloparadox: It's used in `pallets/dactr/src/extensions/check_batch_transactions_mock.rs`.

Basically we cannot impl `WeightToFee` trait for `u64` so we need to create a wrapper first. 
Instead of every mock creating the same wrapper, I just created and placed a default test one in the pallet `lib.rs` so that it can be used everywhere.

Leouarz (approved):

PR #496, opened 2024-03-25
Removed length multiplier & unforked transaction_payment pallet

ToufeeqP:


Leouarz (approved):
Feels great to unfork

PR #514, opened 2024-04-02
Fix query_proof RPC

ToufeeqP:
# Pull Request type

- [x] Bugfix

## Description
Kate query_proof RPC was not extending the columns, As a result, we were not able to fetch the proof for extended cells

## Checklist

### DIFF
--- a/e2e/src/main.rs
+++ b/e2e/src/main.rs
@@ -258,6 +258,48 @@ pub mod tests {
 		let dim = Dimensions::new(rows, cols).unwrap();
 		let res = verify(&pp, dim, &commitment, &dcell).unwrap();
 		assert!(res);
+
+		// Fetch & verify extended cell
+		let cell = Cell { row: 1, col: 1 };
+		let cells = Cells::try_from(vec![cell.clone()]).unwrap();
+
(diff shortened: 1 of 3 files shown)


aterentic (approved):

aterentic-ethernal (approved):

fmiguelgarcia (approved):

e2e/src/main.rs
```diff
@@ -258,6 +258,48 @@ pub mod tests {
...
+				ext.commitment.commitment.clone(),
+				ext.commitment.rows,
+				ext.commitment.cols,
+			),
+		};
+
+		let mut content = [0u8; 80];
+		content.copy_from_slice(&actual_proof);
+		let commitment: [u8; 48] = commitment[48..96].try_into().unwrap();
+		let dcell = kate_recovery::data::Cell {
```
fmiguelgarcia, line 296: It looks like we can refactor this `Cell` because we are encoding the scalar to raw bytes some lines above, and then it is decoded inside `verify` function. 
Maybe it could be a new PR after release.
att. @ToufeeqP @jakubcech

PR #583, opened 2024-05-07
E2E tests move

fmiguelgarcia:
# Pull Request type
- [X] Testing

## Description
Move `avail-subxt` examples into **E2E** tests.
It also run them concurrently, using a couple of tools/functions:
- `fn alice_nonce()`, a central point to read/reserve Alice's nonce, so any transaction in tests will be valid independently of the concurrency of its execution.
- `CTC` or `Concurrency Text Control` is a helper to enforce some test to run alone (non concurrently with other ones). It is used when you want to control which TXs are include in a block, and you want to check data at block scope. For instance, if you want to check `data_proof` or `query_proof`, you cannot run your test concurrently with other which could add extra TXs (like `submit_data` or `send_message`), in a way where expected result will be invalidated.

## HAL-01: Missing Maximum Size Validation
:warning:  **NOTE:** Moved to another PR.
~This includes an E2E test to ensure that `failed_send_message` transactions is **bounded**. It reaches the **maximum number of failed `vector::send_message transactions` per block**, so we can determine the maximum number of integers used by `MemeoryTemporaryStorage`.
We use the minimum size `send_message` transaction possible as an `ArbitraryMessage` and `FungibleToken`.
In order to get that minimum size, and produce a maximum number of failed TXs, the parameters of the call are:~
- ~`message`: an `ArbitraryMessage` with empty data.,~
- ~`to`: Bob's address,~
- ~`domain`: It is just `0` (invalid domain), forcing the TX fail.~

~Having that, the **maximum number of TXs per block is just `136`**, filling the 86% of the block weight. 
That  generates a post inherent TX `failed_send_message` of `136` items (`Compact<u32>`) with an encoded size of  `432 bytes``, and an extra block weight of 12M (**around 0.62 % of total block**).~

### In Summary,
~The worst case means, **`136 * size_of<u32> ~= 544 bytes`  extra memory requirements per block**.
The impact of this post inherent TX **could be up to `0.62%` of the total block weight**. Keep in mind that transactions from regular users could fill up to 85% of the block.~

## Checklist
- [X] I have performed a self-review of my own code.
- [X] The tests pass successfully with `cargo test`.
- [X] The code was formatted with `cargo fmt`.
- [X] The code compiles with no new warnings with `cargo build --release` and `cargo build --release --features runtime-benchmarks`.
- [X] The code has no new warnings when using `cargo clippy`.
- [X] If this change affects documented features or needs new documentation, I have created a PR with a [documentation update](https://github.com/availproject/availproject.github.io).


Leouarz (approved):
maybe e2e_kate.sh could just be renamed e2e.
Nice optimization on the concurrency.

LGTM

fmiguelgarcia:
> e2e_kate.sh could just be renam

fmiguelgarcia:
> maybe e2e_kate.sh could just be renamed e2e. 

You're right!. Done. Thanks @Leouarz

PR #585, opened 2024-05-13
Fix padding on rlp value

0xSasaPrsic:
# Pull Request type

- [ ] Feature
- [x] Bugfix
- [ ] Refactor
- [ ] Format
- [ ] Documentation
- [ ] Testing
- [ ] Other:

## Description
Removes `H256::from_slice` which panics if a provided slice is not 32 bytes.
It appends leading zeros up to 32 bytes if the rlp value (message hash) has leading zeros that are trimmed.  
Prevents code from panicking and gets the correct hash (32 bytes) from the storage proof/account. 

## Checklist


luka-ethernal (approved):
LTGM

pallets/vector/src/storage_utils.rs
```diff
@@ -81,19 +84,42 @@ pub fn get_storage_root(
...
 		Err(StorageError::StorageValueError)
 	}
 }
 
+/// padded_value it converts rlp value to H256 which requires 32 bytes and
+/// appends leading zeros if value is less than 32 bytes, otherwise throws an error.
+fn padded_value(rlp_value: &[u8]) -> Result<H256, StorageError> {
+	let mut slot_value = [0u8; 32].to_vec();
+
+	if rlp_value.len() < 32 {
```
ToufeeqP, line 100: may consider using a match here

ToufeeqP (approved):

pallets/vector/src/storage_utils.rs
0xSasaPrsic: clippy has a suggestion for this but seems more complex than `if` `else`

ToufeeqP: Should be very similar current code, anyway, it's your call

Leouarz (approved):

markopoloparadox (approved):

PR #608, opened 2024-07-08
Added avail-deno and avail-rust

markopoloparadox:


avail-rust/src/events.rs
```diff
@@ -0,0 +1,62 @@
+/* use crate::avail;
```
Leouarz, line 1: Should this be removed ?

avail-rust/src/utils/mod.rs
```diff
@@ -0,0 +1,34 @@
+/*
```
Leouarz, line 1: Should this be removed ?

Leouarz (approved):
LGTM, removing avail-subxt will need to be notified way beforehand

ToufeeqP (approved):

PR #653, opened 2024-09-03
Updated phragmen bounds param

ToufeeqP:


runtime/src/constants.rs
```diff
@@ -286,9 +281,9 @@ pub mod staking {
 		// Note: the EPM in this runtime runs the election on-chain. The election bounds must be
 		// carefully set so that an election round fits in one block.
 		pub ElectionBoundsMultiPhase: ElectionBounds = ElectionBoundsBuilder::default()
-			.voters_count(10_000.into()).targets_count(1_500.into()).build();
+			.voters_count(MaxElectingVoters::get().into()).build();
 		pub ElectionBoundsOnChain: ElectionBounds = ElectionBoundsBuilder::default()
-			.voters_count(5_000.into()).targets_count(1_250.into()).build();
+			.voters_count(5_000.into()).build();
```
fmiguelgarcia, line 286: Could we also define `5_000` as a constant too?

runtime/src/impls.rs
```diff
@@ -455,7 +455,7 @@ impl pallet_election_provider_multi_phase::Config for Runtime {
 	type Fallback = onchain::OnChainExecution<OnChainSeqPhragmen>;
 	type ForceOrigin = EnsureRoot<AccountId>;
 	type GovernanceFallback = onchain::OnChainExecution<OnChainSeqPhragmen>;
-	type MaxWinners = constants::MaxActiveValidators;
+	type MaxWinners = constants::staking::MaxActiveValidators;
```
fmiguelgarcia, line 458: Could we remove `constants::MaxActiveValidators` if that is not used anymore?

ToufeeqP: Its already removed

runtime/src/constants.rs
ToufeeqP: yeah, will do that

markopoloparadox (approved):

fmiguelgarcia (approved):

PR #665, opened 2024-09-19
Fusion mvp

Leouarz:
# Pull Request type

- [x] Feature


pallets/fusion/Cargo.toml
```diff
@@ -0,0 +1,57 @@
...
+[package.metadata.docs.rs]
+targets = ["x86_64-unknown-linux-gnu"]
+
+[dependencies]
+frame-system = { workspace = true, default-features = false }
+hex-literal = { workspace = true }
+log.workspace = true
+
+# Substrate
+codec = { package = "parity-scale-codec", version = "3.0.0", default-features = false, features = ["derive"] }
```
markopoloparadox, line 15: Unrelated: Looks like we don't have this as a workspace dependency.

pallets/fusion/src/lib.rs
```diff
@@ -0,0 +1,1989 @@
...
+		#[pallet::constant]
+		type PalletId: Get<PalletId>;
+
+		/// Maximum allowed for the currency name
+		#[pallet::constant]
+		type MaxCurrencyName: Get<u32>;
+
+		/// Maximum number of members in a pool
+		#[pallet::constant]
+		type MaxMembersPerPool: Get<u32>;
```
markopoloparadox, line 79: Somewhat related: Do we foresee this needing to changeable by TC?

pallets/fusion/src/lib.rs
```diff
@@ -0,0 +1,1989 @@
...
+		_,
+		Blake2_128Concat,
+		EraIndex,
+		Blake2_128Concat,
+		PoolId,
+		FusionExposure<T>,
+		OptionQuery,
+	>;
+
+	/// Stores the fusion era exposure for HistoryDepth eras
```
markopoloparadox, line 215: The comment is wrong.

pallets/fusion/src/lib.rs
```diff
@@ -0,0 +1,1989 @@
...
+		PoolSet {
+			pool_id: PoolId,
+			apy: Option<Perbill>,
+			state: Option<FusionPoolState>,
+			nominator: Option<Option<T::AccountId>>,
+		},
+		/// Event triggered when a user joins a pool
+		PoolJoined {
+			evm_address: EvmAddress,
+			pool_id: PoolId,
```
markopoloparadox, line 312: For sake of consistency, it could be useful to put `pool_id` as the first member of all events.

pallets/fusion/src/lib.rs
```diff
@@ -0,0 +1,1989 @@
...
+			initial_conversion_rate: BalanceOf<T>,
+		) -> DispatchResult {
+			ensure_root(origin)?;
+
+			ensure!(
+				!FusionCurrencies::<T>::contains_key(currency_id),
+				Error::<T>::CurrencyAlreadyExists
+			);
+
+			ensure!(
```
markopoloparadox, line 520: This might be more readable.
```rust
if currency_id == AVAIL_NAITVE_CURRENCY_ID {
   ensure(...)
}
```

pallets/fusion/src/lib.rs
```diff
@@ -0,0 +1,1989 @@
...
+
+			Ok(())
+		}
+
+		/// Creates a new currency
+		#[pallet::call_index(2)]
+		#[pallet::weight(T::WeightInfo::create_fusion_currency())]
+		pub fn create_fusion_currency(
+			origin: OriginFor<T>,
+			currency_id: CurrencyId,
```
markopoloparadox, line 506: Is there a reason why the caller should be able to define the currency_id? Could be simpler to just fetch it from `FusionCurrencyCount`

pallets/fusion/src/lib.rs
```diff
@@ -0,0 +1,1989 @@
...
+				amount,
+			});
+
+			Ok(())
+		}
+
+		/// Creates a new currency
+		#[pallet::call_index(2)]
+		#[pallet::weight(T::WeightInfo::create_fusion_currency())]
+		pub fn create_fusion_currency(
```
markopoloparadox, line 504: We could shorten the name to `create_currency`.

pallets/fusion/src/lib.rs
```diff
@@ -0,0 +1,1989 @@
...
+			FusionPools::<T>::try_mutate(pool_id, |maybe_pool| {
+				let pool = maybe_pool.as_mut().ok_or(Error::<T>::PoolNotFound)?;
+
+				ensure!(
+					pool.state != FusionPoolState::Destroying,
+					Error::<T>::PoolIsDestroying
+				);
+
+				if let Some(apy_value) = apy {
+					pool.apy = apy_value;
```
markopoloparadox, line 755: `pool.apy = api.unwrap_or_else(|| pool.apy)`

pallets/fusion/src/lib.rs
```diff
@@ -0,0 +1,1989 @@
...
+				Ok(())
+			})
+		}
+
+		/// Deletes a pool
+		/// Called once to set the pool to destroying
+		/// Called a second time when everything is cleaned to actually destroy it
+		#[pallet::call_index(8)]
+		#[pallet::weight(T::WeightInfo::create_fusion_currency())]
+		pub fn delete_fusion_pool(origin: OriginFor<T>, pool_id: PoolId) -> DispatchResult {
```
markopoloparadox, line 794: TODO

pallets/fusion/src/lib.rs
```diff
@@ -0,0 +1,1989 @@
...
+			} else {
+				Some(ensure_signed(origin)?)
+			};
+
+			// Fetch the pool and ensure it exists
+			FusionPools::<T>::try_mutate(pool_id, |pool_opt| -> DispatchResult {
+				let pool = pool_opt.as_mut().ok_or(Error::<T>::PoolNotFound)?;
+
+				// If the caller is not root, ensure it's the nominator of the pool
+				if let Some(caller) = who {
```
markopoloparadox, line 832: better naming is needed.

pallets/staking/src/lib.rs
```diff
@@ -0,0 +1,1256 @@
+// This file is part of Substrate.
```
markopoloparadox, line 1: We should move staking back to polkadot-sdk

pallets/fusion/src/lib.rs
Leouarz: Probably not, still changeable by runtime upgrade but we will see.
Pls resolve this if you have not more question

pallets/fusion/src/lib.rs
Leouarz: No reason in particular except simplicity and chosing relevant id compared to existing stuff on EVM
You think it should be changed ?

pallets/fusion/src/lib.rs
Leouarz: What do you mean ?

pallets/fusion/src/lib.rs
Leouarz: What do you mean ?

pallets/staking/src/lib.rs
Leouarz: This will be done at the end to simplify dev

pallets/fusion/src/lib.rs
markopoloparadox: It's a message for me to think about the naming scheme. `Remove` instead of `delete` might be a better name but I will need to do some research and see what nomenclature is used in polkadot sdk.

pallets/fusion/src/lib.rs
markopoloparadox: It's a bit unclear what's going on here. 
```rust
let is_root = ensure_root(origin.clone()).is_ok();
let who = if is_root {
   None
} else {
    Some(ensure_signed(origin)?)
};

if let Some(caller) = who {
    ensure!(Some(&caller) == pool.nominator.as_ref(), Error::<T>::NotAuthorized);
}			
```
I will come up with a better option what's I am done reviewing

pallets/fusion/src/lib.rs
markopoloparadox: If there is a need for it to be user defined then it's fine, otherwise I don't see why we would bother the user to define it

pallets/fusion/src/lib.rs
```diff
@@ -0,0 +1,1989 @@
...
+					);
+				}
+
+				// We cannot change nominations if the pool is destroying
+				ensure!(
+					pool.state != FusionPoolState::Destroying,
+					Error::<T>::PoolIsDestroying
+				);
+
+				// We cannot put 0 nominations if the pool is working
```
markopoloparadox, line 845: What's wrong with having an active pool without targets?

Leouarz: Good question

pallets/fusion/src/lib.rs
```diff
@@ -0,0 +1,1989 @@
...
+		/// Nominates a list of validators for a given pool.
+		#[pallet::call_index(9)]
+		#[pallet::weight(T::WeightInfo::create_fusion_currency())]
+		pub fn nominate(
+			origin: OriginFor<T>,
+			pool_id: PoolId,
+			targets: BoundedVec<T::AccountId, T::MaxTargets>,
+		) -> DispatchResult {
+			// Check if the origin is root, if not, check if it's a signed origin.
+			let is_root = ensure_root(origin.clone()).is_ok();
```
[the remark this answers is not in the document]
Leouarz: Nominator for a pool is optional, for all regular pools, we use root, else the guy need to be nominator

pallets/fusion/src/lib.rs
Leouarz: root can do whatever, if it's a regular address, only nominator can change nominations

pallets/fusion/src/lib.rs
```diff
@@ -0,0 +1,1989 @@
...
+				.ok_or(Error::<T>::UserNotFoundInExposure)?;
+
+			// Ensure the user has not already claimed the reward for this era and pool
+			ensure!(
+				!ClaimedRewards::<T>::contains_key((evm_address, pool_id, era)),
+				Error::<T>::AlreadyClaimed
+			);
+
+			// Calculate the reward ratio
+			let user_share = Pallet::<T>::u256(*user_points);
```
markopoloparadox, line 1593: Isn't `U256::from(*user_points)` shorter than what we have now?

pallets/fusion/src/lib.rs
```diff
@@ -0,0 +1,1989 @@
...
+	}
+
+	/// Clean history depth storages and send old pending rewards to 'RewardRemainder'
+	fn clean_history_depth_storages(era: EraIndex) -> DispatchResult {
+		let history_depth = T::HistoryDepth::get();
+
+		if let Some(era_to_clear) = era.checked_sub(history_depth) {
+			// Clean fusion exposures
+			let pool_count = FusionPoolCount::<T>::get();
+			let _ = FusionExposures::<T>::clear_prefix(era_to_clear, pool_count, None);
```
markopoloparadox, line 1229: Here we don't need to use `FusionPoolCount` as the limit. We can just put `MAX` because at any given time we are not expecting to have more than 1000 pools or something like that.

pallets/fusion/src/lib.rs
```diff
@@ -0,0 +1,1989 @@
...
+		}
+
+		Ok(())
+	}
+
+	/// Compute rewards for each pool and set them in storage
+	/// Exposure was set at the end of the era N for era N
+	/// Reward computatation is done at the end of era N for era N-1
+	fn compute_era_rewards(era: EraIndex, era_duration: u64) -> DispatchResult {
+		if let Some(era_to_process) = era.checked_sub(1) {
```
markopoloparadox, line 1275: Too many indentations makes it hard to reason with it.
For this line you can do the following
```rust
let Some(era_to_process) = era.checked_sub(1) else {
    return Ok(());
};
```

pallets/fusion/src/lib.rs
```diff
@@ -0,0 +1,1989 @@
...
+			let mut total_rewarded_external_pools = BalanceOf::<T>::zero();
+			let mut pool_rewarded: Vec<PoolId> = vec![];
+			let pallet_accounts = Self::accounts();
+			let pallet_account = pallet_accounts.funds_reward_account;
+			let pallet_account_free_balance = T::Currency::free_balance(&pallet_account);
+			let existential_deposit = T::Currency::minimum_balance();
+			let destination_account = pallet_accounts.claimable_reward_account;
+			for (pool_id, fusion_exposure) in FusionExposures::<T>::iter_prefix(era_to_process) {
+				// TODO - check for valid nominations here
+				if let Some(mut pool) = FusionPools::<T>::get(pool_id) {
```
markopoloparadox, line 1286:
```rust
let mut pool = match FusionPools::<T>::get(pool_id) {
  Some(p) => p,
  None => {
	log::error!(
		"🚨 Pool with PoolId {:?} not found for Era {:?}. Reward could not have been set. 🚨",
		pool_id,
		era_to_process
	);
	continue;
  },
};
```

pallets/fusion/src/lib.rs
```diff
@@ -0,0 +1,1989 @@
...
+					if fusion_exposure.total_avail > 0u32.into()
```
markopoloparadox, line 1287:
```rust
if fusion_exposure.total_avail == 0u32.into()
  || pool.members.is_empty()
  || pool.is_paused()
  || pool.is_destroying()
  || Perbill::is_zero(&pool.apy)
{
  continue;
}
```

pallets/fusion/src/lib.rs
Leouarz: Since it's MVP, let's leave it like this then maybe change it after demo'ing it

pallets/fusion/src/lib.rs
Leouarz: We have FusionPoolCount so why not use it ? You wanna save a read ?

pallets/fusion/src/lib.rs
Leouarz: Also, it allows us to create a currency with an id and in the same batch nominating from it and setting it top open instead of waiting

pallets/fusion/src/lib.rs
markopoloparadox: That would be great :)

pallets/fusion/src/lib.rs
Leouarz: Actually should be destroy i guess

pallets/fusion/src/lib.rs
```diff
@@ -0,0 +1,2290 @@
...
+	/// Stores the number of currencies created
+	#[pallet::storage]
+	#[pallet::getter(fn currency_count)]
+	pub type FusionCurrencyCount<T> = StorageValue<_, u32, ValueQuery>;
+
+	/// Stores all the fusion pools
+	#[pallet::storage]
+	#[pallet::getter(fn fusion_pools)]
+	pub type FusionPools<T: Config> =
+		StorageMap<_, Blake2_128Concat, PoolId, FusionPool<T>, OptionQuery>;
```
ToufeeqP, line 145: We can use a non-cryptographic hasher here. This applies to many other storage as well

pallets/fusion/src/lib.rs
```diff
@@ -0,0 +1,2290 @@
...
+			ensure_signed(origin)?;
+			Self::do_dummy_slash(who, bonded_amount, slashed_amount)?;
+			Ok(())
+		}
+
+		/// Pauses the entire pallet, halts rewards creation. Claiming, unbonding and withdrawing is still possible
+		#[pallet::call_index(0)]
+		#[pallet::weight(T::WeightInfo::create_currency())]
+		pub fn pause(origin: OriginFor<T>) -> DispatchResult {
+			ensure_root(origin)?;
```
ToufeeqP, line 532: It is better to replace `ensure_root` in the pallet with `ApprovedOrigin` & configure it in the runtime

pallets/fusion/src/types.rs
```diff
@@ -0,0 +1,368 @@
...
+pub type Points = u128;
+
+/// Type of the currency id
+pub type CurrencyId = u32;
+
+/// Type of the pool id
+pub type PoolId = u32;
+
+/// Prefix used for storing accounts
+pub const FUNDS_ACCOUNT_PREFIX: &str = "funds";
```
ToufeeqP, line 24: In favour of using Enum for all AccountPrefixes, similar to how it's done in nom pools. Any specific advantage of using str here?

ToufeeqP: [ref](https://github.com/availproject/polkadot-sdk/blob/09254eb9f8670f942c68914ff6fc0ffb503aaf23/substrate/frame/nomination-pools/src/lib.rs#L451)

pallets/fusion/src/lib.rs
```diff
@@ -0,0 +1,3044 @@
...
+	/// Stores the total value locked in avail and the maximum total value locked authorized
+	#[pallet::storage]
+	#[pallet::getter(fn tvl_data)]
+	pub type TotalValueLockedData<T> = StorageValue<_, TVLData<T>, ValueQuery>;
+
+	/// Stores all the fusion currencies
+	#[pallet::storage]
+	#[pallet::getter(fn fusion_currencies)]
+	pub type FusionCurrencies<T: Config> =
+		StorageMap<_, Twox64Concat, CurrencyId, FusionCurrency<T>, OptionQuery>;
```
ToufeeqP, line 133: value contains the key itself, if not necessary, we should avoid it.

pallets/fusion/src/lib.rs
```diff
@@ -0,0 +1,3044 @@
...
+
+	/// Stores all the fusion pools
+	#[pallet::storage]
+	#[pallet::getter(fn fusion_pools)]
+	pub type FusionPools<T: Config> =
+		StorageMap<_, Twox64Concat, PoolId, FusionPool<T>, OptionQuery>;
```
ToufeeqP, line 139: same applies here as well, FusionPool has PoolId

pallets/fusion/src/lib.rs
```diff
@@ -0,0 +1,3044 @@
...
+
+	/// Mapping from the pools funds account address to the pool id
+	#[pallet::storage]
+	#[pallet::getter(fn fusion_pool_account_to_id)]
+	pub type FusionPoolsAccountToId<T: Config> =
```
ToufeeqP, line 144: Maybe we can avoid `Fusion` prefixes for most storages when you do cleaning next time?

pallets/fusion/src/lib.rs
```diff
@@ -0,0 +1,3044 @@
...
+		CurrencyId,
+		BalanceOf<T>,
+		OptionQuery,
+	>;
+
+	/// Stores the next currency changes to be applied next era
+	#[pallet::storage]
+	#[pallet::getter(fn fusion_currency_rate_changes)]
+	pub type FusionCurrencyRateChanges<T: Config> =
+		StorageMap<_, Twox64Concat, CurrencyId, BalanceOf<T>, OptionQuery>;
```
ToufeeqP, line 204: Can make it ValueQuery

pallets/fusion/src/lib.rs
```diff
@@ -0,0 +1,3044 @@
...
+
+	/// Mapping from EVM Address to Substrate address
+	#[pallet::storage]
+	#[pallet::getter(fn fusion_evm_to_substrate_address)]
+	pub type FusionEVMToSubstrateAddress<T: Config> =
```
ToufeeqP, line 209: Is it better to replace EvmAddress with FusionAddress everywhere?

pallets/fusion/src/lib.rs
```diff
@@ -0,0 +1,3044 @@
...
+		StorageMap<_, Blake2_128Concat, EvmAddress, T::AccountId, OptionQuery>;
+
+	#[pallet::storage]
+	#[pallet::getter(fn era_durations)]
+	pub type EraDurations<T: Config> = StorageMap<_, Twox64Concat, EraIndex, u64, OptionQuery>;
```
ToufeeqP, line 214: Could we pls add a line of doc here for clarity?

pallets/fusion/src/lib.rs
```diff
@@ -0,0 +1,3044 @@
...
+		Twox64Concat,
+		PoolId,
+		FusionExposure<T>,
+		OptionQuery,
+	>;
+
+	/// Store the pools that backed the validator set as the key
+	#[pallet::storage]
+	#[pallet::getter(fn fusion_pools_from_validator)]
+	pub type FusionPoolsFromValidator<T: Config> = StorageDoubleMap<
```
ToufeeqP, line 232: can we rename this to more meaningful like `ValidatorBackingPools` or something else, little confusing to me right now? Also, `set` in the doc is confusing

pallets/fusion/src/lib.rs
```diff
@@ -0,0 +1,3044 @@
...
+	/// Used mainly to quickly determine if a slashed nominator is from Fusion pallet
+	#[pallet::storage]
+	#[pallet::getter(fn has_pending_slash)]
+	pub type HasPendingSlash<T: Config> = StorageDoubleMap<
+		_,
+		Twox64Concat,
+		EraIndex,
+		Twox64Concat,
+		(T::AccountId, T::AccountId),
+		bool,
```
ToufeeqP, line 272: Having an entry for the pool account itself indicates the pool's slashing exposure. Instead of bool, we can store the slashable balance, similar to the info in the `ValidatorSlashInEra` storage of the staking pallet.

pallets/fusion/src/lib.rs
Leouarz: Done but used the opposite for clarity: PoolsBackingValidator

jakubcech:
lgtm so far

pallets/fusion/src/mock.rs
```diff
@@ -0,0 +1,84 @@
...
+}
+
+#[derive_impl(pallet_balances::config_preludes::TestDefaultConfig as pallet_balances::DefaultConfig)]
+impl pallet_balances::Config for Test {
+	type AccountStore = System;
+}
+
+parameter_types! {
+	pub const FusionPalletId: PalletId = PalletId(*b"avl/fusi");
+	pub const MaxCurrencyName: u32 = 32;
```
markopoloparadox, line 45: What do you think aboout `MaxCurrencyNameLength`?

pallets/fusion/src/lib.rs
```diff
@@ -0,0 +1,3085 @@
...
+			ensure_root(origin)?;
+
+			ensure!(
+				!Currencies::<T>::contains_key(currency_id),
+				Error::<T>::CurrencyAlreadyExists
+			);
+
+			ensure!(name.len() > 0, Error::<T>::InvalidName);
+			ensure!(nb_decimals > 0, Error::<T>::InvalidNumberOfDecimals);
+			ensure!(max_amount > 0, Error::<T>::InvalidMaxNumber);
```
markopoloparadox, line 658: Do we need to check that `min_amount` needs to be less than `max_amount`?

pallets/fusion/src/lib.rs
```diff
@@ -0,0 +1,3085 @@
...
+		) -> DispatchResult {
+			ensure_root(origin)?;
+
+			ensure!(
+				!Currencies::<T>::contains_key(currency_id),
+				Error::<T>::CurrencyAlreadyExists
+			);
+
+			ensure!(name.len() > 0, Error::<T>::InvalidName);
+			ensure!(nb_decimals > 0, Error::<T>::InvalidNumberOfDecimals);
```
markopoloparadox, line 657: Is this a requirement? Is there something wrong with having currencies without decimal places?

Leouarz: There is not currency without decimals place in Crypto. I would have liked to say something like, minimum amount of decimals is 4 or 8, in the mean time, greater than zero is fine

PR #718, opened 2025-01-31
fix: typos in documentation files

kilavvy:


markopoloparadox:
Hello,

The documentation that you have fixed will not have an effect as it will be changed back once we regenerate the api_dev file. As this is upstream issue, I would suggest you to visit the official [polkadot-sdk](https://github.com/paritytech/polkadot-sdk) github and fix the typo there.

PR #720, opened 2025-02-12
Add Transaction State (Status) RPC

markopoloparadox:
Script to test this feature:
```bash
#!/bin/bash

# Check if the transaction TX argument is provided
if [ -z "$1" ]; then
    echo "Usage: $0 <transaction_tx>"
    exit 1
fi

transaction_tx="$1"

curl -H "Content-Type: application/json" -d "{
  \"jsonrpc\": \"2.0\",
  \"method\": \"transaction_state\",
  \"params\": [\"$transaction_tx\"],
  \"id\": 0
}" http://127.0.0.1:9944
echo ""
```

### DIFF
--- a/node/src/rpc.rs
+++ b/node/src/rpc.rs
@@ -94,14 +94,18 @@ pub struct FullDeps<C, P, SC, B> {
 	pub babe: BabeDeps,
 	/// GRANDPA specific dependencies.
 	pub grandpa: GrandpaDeps<B>,
-	/// The maximum number of cells that can be requested in one go.
-	pub kate_max_cells_size: usize,
-	/// Enable Kate RPCs
-	pub kate_rpc_enabled: bool,
-	/// Enable Kate RPCs Metrics
+	/// Kate RPC specific dependencies.
 	///
-	/// Should not be used unless unless you know what you're doing.
-	pub kate_rpc_metrics_enabled: bool,
+	/// Available configs:
+	/// - pub max_cells_size: usize,
+	/// - pub rpc_enabled: bool,
+	/// - pub rpc_metrics_enabled: bool,
+	pub kate_rpc_deps: kate_rpc::Deps,
+	/// Transaction RPC specific dependencies.
+	///
+	/// Available configs:
+	/// - TxStateSender,
+	pub transaction_rpc_deps: Option<transaction_rpc::Deps>,
 }
 
 /// Instantiate all Full RPC extensions.
@@ -152,9 +156,8 @@ where
 		deny_unsafe,
 		babe,
 		grandpa,
-		kate_max_cells_size,
-		kate_rpc_enabled,
-		kate_rpc_metrics_enabled,
+		kate_rpc_deps,
+		transaction_rpc_deps,
 	} = deps;
 
 	let BabeDeps {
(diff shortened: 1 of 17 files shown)


node/src/rpc.rs
```diff
@@ -94,14 +94,10 @@ pub struct FullDeps<C, P, SC, B> {
...
 	pub grandpa: GrandpaDeps<B>,
-	/// The maximum number of cells that can be requested in one go.
-	pub kate_max_cells_size: usize,
-	/// Enable Kate RPCs
-	pub kate_rpc_enabled: bool,
-	/// Enable Kate RPCs Metrics
-	///
-	/// Should not be used unless unless you know what you're doing.
-	pub kate_rpc_metrics_enabled: bool,
+	/// TODO
```
prabal-banerjee, line 97: Can we resolve this todo and just mention the configs which are wrapped inside it?

Leouarz (approved):

rpc/kate-rpc/src/lib.rs
```diff
@@ -30,6 +30,18 @@ pub type Cells = BoundedVec<Cell, MaxCells>;
 
 pub mod metrics;
 
+#[derive(Clone, Default)]
+pub struct Deps {
```
ToufeeqP, line 34: Thanks for grouping them :)

rpc/system-rpc/src/lib.rs
```diff
@@ -0,0 +1,82 @@
...
+}
+
+pub type OneShotTxStateSender = oneshot::Sender<Vec<TransactionState>>;
+pub type TxStateReceiver = Receiver<(H256, bool, OneShotTxStateSender)>;
+pub type TxStateSender = Sender<(H256, bool, OneShotTxStateSender)>;
+pub type TxStateChannel = (H256, bool, OneShotTxStateSender);
+
+#[rpc(client, server)]
+pub trait TransactionState {
+	#[method(name = "system_transaction_state")]
```
ToufeeqP, line 38: method name part should be camelCase as we're following this standard across all RPC's, also can we change the grouping of this RPC from system to something else to differentiate between default ones & our custom RPC's?

runtime/src/apis.rs
```diff
@@ -237,6 +239,36 @@ impl_runtime_apis! {
 	}
 
 
+	impl frame_system_rpc_runtime_api::SystemEventsApi<Block> for Runtime {
+		fn fetch_transaction_success_status() -> Vec<frame_system_rpc_runtime_api::TransactionSuccessStatus> {
+			use frame_system_rpc_runtime_api::TransactionSuccessStatus;
+			use frame_system::Event;
+
+			let mut results: Vec<TransactionSuccessStatus> = Vec::new();
+			let event_records = System::read_events_no_consensus();
```
ToufeeqP, line 248: Have you tested how much memory & time it takes for large block with large number of events? eg: with remark_with_events calls

node/src/rpc.rs
markopoloparadox: Done. Thank you

rpc/system-rpc/src/lib.rs
markopoloparadox: Done. Thank you

runtime/src/apis.rs
markopoloparadox: As discussed in the call, the feature is by default disabled and once it deployed to our testnet we will run some benchmarks to see how "bad" it can get.

ToufeeqP (approved):

PR #728, opened 2025-02-27
Add binary signing to the releaser

jakubcech:
# Add GPG signature verification to release binaries

## Summary
This PR changes the release process by implementing GPG signing for all **binary** artifacts. This allows users to cryptographically verify that binaries were built by our official CI process and haven't been tampered with.

## Changes
- Added a new `setup_gpg` job that imports our signing key and publishes the public key as a release asset
- Modified all build jobs in the releaser to:
  - Generate SHA-256 checksums for each binary
  - Create detached GPG signatures for both binaries and checksums
  - Upload signatures and checksums to GitHub releases alongside binaries
- Added dependency relationships to ensure signing setup completes before build jobs

## Security considerations
- Uses GitHub organization secrets for GPG credentials (key, key ID, and passphrase)
- Pins third-party actions to specific commit hashes for supply chain security (only the one handling keys directly, not the one uploading artifacts because \_O_/)
- Creates a complete verification chain: signed binaries, signed checksums, and published public key

## User impact
Users can now verify the authenticity of our binaries using standard GPG tools. Documentation has been updated with verification instructions.

### DIFF
--- a/.github/workflows/releaser_binaries.yml
+++ b/.github/workflows/releaser_binaries.yml
@@ -8,8 +8,43 @@ on:
   workflow_dispatch:
 
 jobs:
+  # Create a job to import and publish the GPG public key
+  setup_gpg:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Import GPG key
+        id: import_gpg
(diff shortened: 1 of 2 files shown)


markopoloparadox:
LGTM

markopoloparadox (approved):

PR #754, opened 2025-06-20
Implement system fetch events api rpc

markopoloparadox:


rpc/kate-rpc/src/system.rs
```diff
@@ -0,0 +1,87 @@
...
+}
+
+/// Error type for this RPC API.
+pub enum Error {
+	/// Generic runtime error.
+	RuntimeApiError,
+}
+
+impl Error {
+	pub fn runtime_api_error<'a>(msg: String) -> ErrorObject<'a> {
```
ToufeeqP, line 54: Since the function is returning an `ErrorObject`, I think it's better to name it likewise & make it an associated function
```suggestion
	pub fn into_error_object<'a>(self, msg: String) -> ErrorObject<'a> {
		ErrorObject::owned(self.into(), msg, None::<()>)
       }
```

runtime/src/apis.rs
```diff
@@ -579,3 +621,73 @@ impl_runtime_apis! {
 		}
 	}
 }
+
+fn decode_runtime_event_v1(event: &super::RuntimeEvent) -> Option<Vec<u8>> {
+	use super::*;
+	use codec::Encode;
+
+	match event {
+		RuntimeEvent::Sudo(e) => match e {
```
ToufeeqP, line 630: So, filtering is possible only on this subset of events? Dont we want to add bridge events to this set?

rpc/kate-rpc/src/system.rs
markopoloparadox: You are right, thank you!

ToufeeqP (approved):

PR #762, opened 2025-07-22
Optimize fetch extrinsic rpc

markopoloparadox:


rpc/kate-rpc/src/system.rs
```diff
@@ -217,6 +217,8 @@ where
 			}
 		}
 
+		cache.promote_block(block_hash);
```
ToufeeqP, line 220: I think we can simply use the LRU cache instead of implementing our own custom cache

ToufeeqP (approved):

PR #800, opened 2026-02-17
Feat/sidecar

vbhattaccmu:
# Pull Request type

- [x] Feature
- [ ] Bugfix
- [ ] Refactor
- [ ] Format
- [ ] Documentation
- [ ] Testing
- [ ] Other:

## Checklist

### DIFF
--- a/blob/src/p2p.rs
+++ b/blob/src/p2p.rs
@@ -86,7 +86,8 @@ pub fn get_blob_p2p_config<B: BlockT, N: NetworkBackend<B, <B as BlockT>::Hash>>
 	)
 }
 
-pub type EvalClaimsCache = Arc<Mutex<HashMap<BlobHash, EvalClaimsMessage>>>;
+/// Cache key for eval claims: (block_hash, blob_hash).
+pub type EvalClaimsCache = Arc<Mutex<HashMap<(H256, BlobHash), EvalClaimsMessage>>>;
 
 static GLOBAL_EVAL_SENDER: once_cell::sync::OnceCell<async_channel::Sender<EvalClaimsMessage>> =
 	once_cell::sync::OnceCell::new();
@@ -291,10 +292,14 @@ where
 		});
 	}
 
-	pub fn get_eval_data_for_blob(&self, blob_hash: H256) -> Option<EvalClaimsMessage> {
+	pub fn get_eval_data_for_blob(
+		&self,
+		block_hash: H256,
+		blob_hash: H256,
+	) -> Option<EvalClaimsMessage> {
 		self.eval_claims_cache
 			.as_ref()
-			.and_then(|c| c.lock().get(&blob_hash).cloned())
+			.and_then(|c| c.lock().get(&(block_hash, blob_hash)).cloned())
 	}
 
 	/// Light node: subscribe to eval_claims topic and cache messages for verification against header.
@@ -334,7 +339,14 @@ where
 							if let Ok(msg) =
 								EvalClaimsMessage::decode(&mut &notification.message[..])
 							{
-								cache.lock().insert(msg.blob_hash, msg);
+								log::info!(
+									target: LOG_TARGET,
+									"Received eval claims message for block {:?}, blob {:?}, app_id {:?}",
+									msg.block_hash,
+									msg.blob_hash,
+									msg.app_id
+								);
+								cache.lock().insert((msg.block_hash, msg.blob_hash), msg);
 							}
 						}
 					}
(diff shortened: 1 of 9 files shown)


blob/src/p2p.rs
```diff
@@ -196,17 +222,29 @@ where
 			None,
 		);
 
-		let topic = <<Block::Header as HeaderT>::Hashing as HashT>::hash("blob_topic".as_bytes());
-		let incoming_receiver = gossip_engine.messages_for(topic);
+		let blob_topic =
+			<<Block::Header as HeaderT>::Hashing as HashT>::hash("blob_topic".as_bytes());
+		let eval_claims_topic =
+			<<Block::Header as HeaderT>::Hashing as HashT>::hash(EVAL_CLAIMS_TOPIC);
```
Leouarz: Might be good to have 'blob_topic' as a const too, so we have to disparity

blob/src/p2p.rs
```diff
@@ -196,17 +222,29 @@ where
...
+							_ => break,
+						}
+					}
+					maybe_eval = eval_claims_cmd_receiver.recv().fuse() => {
+						match maybe_eval {
+							Ok(msg) => {
+								gossip_engine.gossip_message(eval_claims_topic, msg.encode(), false)
 							},
 							_ => break,
 						}
```
Leouarz, line 257: Would it make sense to have the eval_claims_cmd_receiver in its own spawned task ?

Leouarz (approved):

blob/src/rpc.rs
```diff
@@ -1357,6 +1357,30 @@ pub async fn store_and_gossip_blob(
...
+
+	// Publish eval claim + proof to sidecar topic so light nodes can verify against header
+	if let (Some(seed), Some(claim), Some(proof)) = (
+		blob_metadata.eval_point_seed,
+		blob_metadata.eval_claim.as_ref(),
+		blob_metadata.fri_eval_proof.as_ref(),
+	) {
+		if let Some(eval_sender) = friends.externalities.eval_claims_cmd_sender() {
+			let eval_msg = avail_blob::types::EvalClaimsMessage {
+				block_hash: finalized_block_hash.into(),
```
ToufeeqP: We shouldn't be publishing the eval data during blob processing/gossip, instead it should be done either during block production or import, as the block_hash we refer here is of the block in which the blob is included, and that is available only after we have constructed the block/imported it. The finalised block_hash is for a different purpose, and it can be different for different nodes based on their network at the time of their processing.

blob/src/types.rs
```diff
@@ -407,6 +407,22 @@ impl BlobEvalData {
 	}
 }
 
+/// Message for the eval claims gossipsub topic: per-blob eval claim + proof for light nodes to verify against header.
+/// Replaces carrying this data in the block body; light client can query from topic and verify.
+#[derive(Debug, Clone, PartialEq, Eq, Encode, Decode, TypeInfo, Serialize, Deserialize)]
+pub struct EvalClaimsMessage {
+	/// Block hash this eval data belongs to
+	pub block_hash: H256,
```
ToufeeqP, line 412: Maybe we can use the app_id instead of block_hash, which enables us to publish the eval data during blob processing & also lightnodes can filter whether to validate the eval_proof based on app_id.

ToufeeqP (commented):
@vbhattaccmu Were you able to test light-nodes validating the eval_proof with this setup? We need to change the key definition on the publishing side, or else change the key itselfo

blob/src/p2p.rs
vbhattaccmu: added a separate const for blob_topic in latest commit

blob/src/types.rs
vbhattaccmu: using app_id instead of block_hash in latest commit. Light nodes are caching eval messages keyed by blob_hash and can  filter per app_id.

vbhattaccmu:
> @vbhattaccmu Were you able to test light-nodes validating the eval_proof with this setup? We need to change the key definition on the publishing side, or else change the key itselfo

not yet fully. will need to test

blob/src/rpc.rs
vbhattaccmu: have updated eval data publishing after block import.

blob/src/p2p.rs
vbhattaccmu: added as a spawned task in latest commit

blob/src/types.rs
ToufeeqP: I think you forgot to push it?

PR #803, opened 2026-02-23
Custom rpc cleanup

markopoloparadox:


rpc/kate-rpc/src/system.rs
```diff
@@ -262,159 +281,104 @@ where
...
 			.block_number_from_id(&sp_runtime::generic::BlockId::Hash(hash.into()))
 			.map_err(|err| Error::Other.into_error_object(err.to_string()))?;
 		Ok(result.map(|x| x.into()))
 	}
+
+	async fn fetch_block_timestamp(&self, block_id: types::BlockId) -> RpcResult<u64> {
+		use types::BlockId;
+
+		const TIMESTAMP_SET_PALLET_ID: u8 = 3;
+		const TIMESTAMP_SET_VARIANT_ID: u8 = 3;
```
ToufeeqP: I guess you meant the call_id here, if so it should be 0 not 3 right?

rpc/kate-rpc/src/system.rs
```diff
@@ -433,214 +397,89 @@ pub mod fetch_events_v1 {
...
-	pub struct ExtrinsicInformation {
-		pub encoded: Option<String>,
-		#[cfg_attr(feature = "ts", ts(as = "String"))]
+	pub struct Extrinsic {
+		pub data: String,
 		pub tx_hash: H256,
 		pub tx_index: u32,
 		pub pallet_id: u8,
-		pub call_id: u8,
+		pub variant_id: u8,
```
ToufeeqP, line 545: Any specific reason for changing the field name from call_id to variant_id? call_id is more familiar imo

rpc/kate-rpc/src/system.rs
```diff
@@ -433,214 +397,89 @@ pub mod fetch_events_v1 {
...
+		PalletCall((u8, u8)),
 	}
 
 	#[derive(Default, Clone, Serialize, Deserialize)]
-	#[cfg_attr(feature = "ts", derive(ts_rs::TS))]
-	#[cfg_attr(feature = "ts", ts(export, export_to = "FetchExtrinsics.ts"))]
-	pub struct SignatureFilterOptions {
+	pub struct SignatureFilter {
 		pub ss58_address: Option<String>,
 		pub app_id: Option<u32>,
```
ToufeeqP: Since we've removed the app_id extension, I think we no longer need to filter by app_id

rpc/kate-rpc/src/system.rs
markopoloparadox: variant id is used everywhere in subxt and polkadotjs creates. 
Besides that call id can mean (pallet_id, variant_id) as well which is confusing. Events have the same problem.

rpc/kate-rpc/src/system.rs
markopoloparadox: Yeah that was removed

ToufeeqP (approved):

PR #853, opened 2026-09-08
apply operational extrinsic class cap

ToufeeqP:
# Pull Request type

- [x] Bugfix

## Checklist
- [ ] I have performed a self-review of my own code.
- [ ] The tests pass successfully with `cargo test`.
- [ ] The code was formatted with `cargo fmt`.


runtime/src/constants.rs
```diff
@@ -118,6 +118,7 @@ pub mod system {
 
 	pub const NORMAL_DISPATCH_RATIO_PERBILL: Perbill =
 		Perbill::from_percent(NORMAL_DISPATCH_RATIO as u32);
+	pub const OPERATIONAL_DISPATCH_RATIO_PERBILL: Perbill = Perbill::from_percent(90);
```
Leouarz: Not sure if this is enough, since post inherent extrinsic can reach like 96mb with max blobs, I would recommend going further and keep only like 10 / 15 mb for operational and normal and keep the rest for mandatory. Also this would require a change in the way we compute costs based on weight (maybe not needed)

ToufeeqP: Thanks for pointing this out, will update the blockLengths accordingly. And yeah, I think the existing weight config can stay as-is for this fix because proof gen & verification happens outside runtime execution. I will create a separate PR to benchmark & account for ev_proof decoding time while including the post-inherent in the block weights.

Leouarz (approved):

